PR #32, opened 2020-09-22
Move Construct to Container

m-nash:
Fix issue with resourceIdentifier throwing on resourcegroup constructor


azure-proto-core/ResourceListOperations.cs
```diff
@@ -60,7 +60,7 @@ public class ResourceListOperations
             //if armfilter == null && listbysub exists for the rp use rp call
             //else use arm list call
             ResourceType type;
-            if (!ArmClient.Registry.TryGetResourceType<T>(out type))
+            if (!ArmClient.Registry.TryGetResourceType<U, T>(out type))
```
markcowl, line 63: Per above comments, I think we need an item to remove this dependency on the registry - we could follow several patterns, like always allowing a default constructor of the resource types, or using reflection over a static ResourceType property in each class.

azure-proto-network/Extensions/ResourceGroupExtension.cs
```diff
@@ -0,0 +1,126 @@
...
+using Azure.ResourceManager.Network.Models;
+using azure_proto_core;
+using azure_proto_core.Resources;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace azure_proto_network
+{
+    public static class ResourceGroupExtension
```
markcowl, line 11: Note: I think the class name here should be plural (which the previous one also was not)

markcowl (changes requested):

markcowl:
A few monor change suggestions - for the registry issue, it is fine to add a task or a TODO

azure-proto-compute/Placeholder/PhVirtualMachine.cs
```diff
@@ -10,8 +10,8 @@ public class PhVirtualMachine : TrackedResource<VirtualMachine>, ITaggable
         public static ResourceType ResourceType => "Microsoft.Compute/virtualMachines";
         static PhVirtualMachine()
         {
-            ArmClient.Registry.Register<PhVirtualMachine>(
-               new azure_proto_core.Internal.ArmResourceRegistration<PhVirtualMachine>(
+            ArmClient.Registry.Register<VirtualMachineOperations, PhVirtualMachine>(
```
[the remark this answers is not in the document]
m-nash: for now yes, it will be replaced at some point, but its still the way we get certain information

markcowl (approved):

PR #36, opened 2020-10-15
Finished with edits on resourcelist and moving list to container classes

m-nash:


azure-proto-compute/AvailabilitySetOperations.cs
```diff
@@ -7,21 +7,13 @@
...
-        }
-        public AvailabilitySetOperations(OperationsBase parent, TrackedResource context) : base(parent, context)
-        {
-        }
-        public AvailabilitySetOperations(OperationsBase parent, ResourceIdentifier context) : base(parent, context)
-        {
-        }
+        public AvailabilitySetOperations(ArmClientContext context, TrackedResource resource) : base(context, resource) { }
+
+        public AvailabilitySetOperations(ArmClientContext context, ResourceIdentifier id) : base(context, id) { }
```
markcowl, line 15: Do we still need this overload?

azure-proto-core/ArmClient.cs
```diff
@@ -22,15 +22,23 @@ public class ArmClient
...
         {
-            Registry.Register<ResourceGroupOperations, PhResourceGroup>(
-                new azure_proto_core.Internal.ArmResourceRegistration<ResourceGroupOperations, PhResourceGroup>(
+            Registry.Register(
+                new azure_proto_core.Internal.ArmResourceRegistration<ResourceGroupContainerOperations, SubscriptionOperations, ResourceGroupOperations, PhResourceGroup>(
                     new ResourceType("Microsoft.Resources/resourceGroups"),
                     (o, r) => new ResourceGroupContainerOperations(o, r),
                     (o, r) => new ResourceGroupOperations(o, r.Id)));
+
+            Registry.Register(
```
markcowl, line 31: I know I keep asking this, but what do we continue to use this for?

azure-proto-core/ArmClient.cs
```diff
@@ -186,17 +192,20 @@ public ResourceGroupOperations ResourceGroup(PhResourceGroup resourceGroup)
...
         {
             if (location == null)
             {
                 location = azure_proto_core.Location.Default;
             }
 
-            ResourceContainerOperations<ResourceOperationsBase<T>, T> container;
-            if (!Registry.TryGetContainer(this.ClientContext, new ArmResource($"/subscriptions/{subscription}/resourceGroups/{resourceGroup}", location), out container))
+            TContainer container;
+            if (!Registry.TryGetContainer<TContainer, ArmResource, TOperations, TResource>(this.ClientContext, new ArmResource($"/subscriptions/{subscription}/resourceGroups/{resourceGroup}", location), out container))
```
markcowl, line 206: At this point, do we really need the registry to do this?  Can't we use Activator.CreateInstance to instantiate the Container class and then call the create method with the given arguments?

Feels like we would only need the registry if we wanted to provide only the resource type and arguments.

azure-proto-core/ArmResourceOperations.cs
```diff
@@ -0,0 +1,47 @@
...
+        public ArmResourceOperations(ArmClientContext context, ResourceIdentifier id) : base(context, id) { }
+
+        public override void Validate(ResourceIdentifier identifier)
+        {
+            //the id can be of any type so do nothing
+        }
+
+        public override ArmOperation<ArmResourceOperations> AddTag(string key, string value)
+        {
+            throw new NotImplementedException();
```
markcowl, line 19: We should be able to use the ResourceManagementClient to implement these

markcowl (changes requested):

azure-proto-compute/AvailabilitySetOperations.cs
m-nash: The tracked resource one?  Yes its used by the Create method since you get a Model back and we need to pass that Model in here so its fullly filled out.

azure-proto-core/ArmClient.cs
m-nash: This will be in a future change, just making the edits needed to get this task working.

azure-proto-core/ArmResourceOperations.cs
m-nash: Ill add a TODO

azure-proto-compute/AvailabilitySetOperations.cs
markcowl: No, I mean the ResourceIdentifier overload

markcowl (approved):

PR #40, opened 2020-10-29
Add a mechanism to override individual resource api versions via ArmC…

m-nash:
…lientOptions


azure-proto-compute/VersionOverrides/AvailabilitySetsApiVersions.cs
```diff
@@ -0,0 +1,26 @@
...
+    public class AvailabilitySetsApiVersions : ApiVersionsBase
+    {
+        public static readonly AvailabilitySetsApiVersions V2020_05_01 = new AvailabilitySetsApiVersions("2020-05-01");
+        public static readonly AvailabilitySetsApiVersions V2019_12_01 = new AvailabilitySetsApiVersions("2019-12-01");
+        public static readonly AvailabilitySetsApiVersions Default = V2020_05_01;
+
+        static AvailabilitySetsApiVersions()
+        {
+            _validValues.Add(V2020_05_01);
+            _validValues.Add(V2019_12_01);
```
bquantump, line 14: How is validValues used? I only see added to ever and never checked (contains called). Also, are not the valid values restricted by lines 7-8 since you cannot set AvailabilitySetsApiVersions any other way since the string operator is not implemented and the constructor is private?

azure-proto-core/ArmClientOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace azure_proto_core
+{
+    public class ArmClientOptions
+    {
+        private Dictionary<Type, object> _overrides = new Dictionary<Type, object>();
+        private static readonly object _overridesLock = new object();
+
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public object GetOverrideObject<T>(Func<object> ctor)
```
bquantump, line 13: Does this have to be public? Would this be used outside of extensions?

m-nash: It does have to be public its used by other DLLs as in the example of ArmClientOptionsExtensions.cs in azure-proto-compute

azure-proto-compute/VersionOverrides/AvailabilitySetsApiVersions.cs
m-nash: yeah you are right, after I removed the string -> ApiVersion implicit this is no longer required

azure-proto-core/ArmClientOptions.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            object overrideObject;
+            Type type = typeof(T);
+            if (!_overrides.TryGetValue(type, out overrideObject))
```
bquantump, line 17: Is the purpose of restricting this to only being made once so that we can't have two api versions or something in effect at once? Since, if the type is already made, we never want to make it again, we don't call ctor()

m-nash: correct, this is basically giving us a singleton per instance of ArmClientOptions, but still allows each instance of ArmClientOptions to have different versions

azure-proto-compute/ComputeApiVersions.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace azure_proto_compute
+{
+    public class ComputeRestVersions
```
weshaggard, line 3:
```suggestion
    public class ComputeRestApiVersions
```

azure-proto-core/ApiVersionsBase.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace azure_proto_core
+{
+    public class ApiVersionsBase : IEquatable<string>, IComparable<string>
+    {
+        private string _value;
```
weshaggard, line 8:
```suggestion
        private readonly string _value;
```

azure-proto-core/ApiVersionsBase.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public override string ToString()
+        {
+            return _value;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is ApiVersionsBase)
+                return Equals(obj as ApiVersionsBase);
```
weshaggard, line 46: I don't see an overload of Equals overload that take an ApiVersionBase. What method does this end up calling? I'm not sure if it recursively calls this method or if it calls the string overload?

azure-proto-compute/ComputeApiVersions.cs
weshaggard: Or perhaps just matching the file name and calling this type `ComputeApiVersions`

azure-proto-compute/VersionOverrides/AvailabilitySetsApiVersions.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    public class AvailabilitySetsApiVersions : ApiVersionsBase
+    {
+        public static readonly AvailabilitySetsApiVersions V2020_05_01 = new AvailabilitySetsApiVersions("2020-05-01");
+        public static readonly AvailabilitySetsApiVersions V2019_12_01 = new AvailabilitySetsApiVersions("2019-12-01");
+        public static readonly AvailabilitySetsApiVersions Default = V2020_05_01;
+
+        private AvailabilitySetsApiVersions(string value) : base(value) { }
+
+        public static implicit operator string(AvailabilitySetsApiVersions version)
```
weshaggard, line 13: Is this needed? Does the operator on the base get picked up?

azure-proto-core/ApiVersionsBase.cs
```diff
@@ -0,0 +1,72 @@
...
+            if (other == null)
+                return false;
+            return other == _value;
+        }
+
+        public int CompareTo(string other)
+        {
+            if (other == null)
+                return 1;
+            return DateTime.Parse(_value).CompareTo(DateTime.Parse(other));
```
weshaggard, line 69: Isn't this baking in too much knowledge about the string? What if we end up having something like "10-10-2020-2" or something?

azure-proto-core/ApiVersionsBase.cs
m-nash: It calls the string version using the implicit.  This avoids ambiguous method error if a consumer uses
ApiVersionBase version;
versioin.Equals(null);

azure-proto-compute/VersionOverrides/AvailabilitySetsApiVersions.cs
m-nash: I couldn't find a way to get implicit to be inherited from base, if you know of a way I would prefer that, but from everything i could find you must implement it in each concrete class.  This will be auto generated so shouldn't be a big deal

azure-proto-core/ApiVersionsBase.cs
m-nash: if we change the format of the string we will definitely need to change this implementation.  Given we are pretty set on calver I don't see this as a risk.

azure-proto-core/ApiVersionsBase.cs
weshaggard: Perhaps we should make this explicit and do `Equals(obj as string)` here as well.

m-nash: I do have that

            if (obj is ApiVersionsBase)
                return Equals(obj as ApiVersionsBase);
            if (obj is string)
                return Equals(obj as string);

azure-proto-core/ApiVersionsBase.cs
```diff
@@ -66,7 +68,32 @@ public int CompareTo(string other)
...
                 return 1;
-            return DateTime.Parse(_value).CompareTo(DateTime.Parse(other));
+
+            Match otherMatch = Regex.Match(other, RegPattern);
+            Match thisMatch = Regex.Match(_value, RegPattern);
+
+            string otherDatePart = otherMatch.Groups[1].Value;
+            string thisDatePart = thisMatch.Groups[1].Value;
+
+            if(otherDatePart == thisDatePart)
```
weshaggard, line 78: nit:
```suggestion
            if (otherDatePart == thisDatePart)
```

azure-proto-core/ArmClientOptions.cs
bquantump: Ic, that makes sense since it has to be use in ArmClientOptionsExtensions like classes which need to be static given our design. Kind of rules out inheritance or anything like that.

bquantump (approved):

weshaggard (approved):

PR #44, opened 2020-11-10
Add SystemAssignedIdentity.cs and its tests

nisha-bhatia:


m-nash (changes requested):

azure-proto-core-test/IdentityTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            Assert.AreEqual(identity1.Equals(identity2), true);
+        }
+
+        [TestCase]
+        public void EqualsMethodFalse()
+        {
+            Identity identity1 = new Identity();
+            Identity identity2 = new Identity();
+
+            identity1.ResourceId = "/subscriptions/6b085460-5f21-477e-ba44-1035046e9101/resourceGroups/nbhatia_test/providers/Microsoft.Web/sites/autoreport";
```
markcowl, line 88: In general, it makes sense to think about the equivalency classes in inputs like equal and non-equal identities, for example
(0) identities that are null or empty (if empty is possible)
(1) identities that consist of exactly the same characters
(2) identities that are equivalent but not identical (identities that vary by capitalization)

and make sure we have tests for these, irrespective of code coverage - this way we have regression protection for important scenarios (like case-insensitive comparison), regardless of the implementation details

azure-proto-core-test/IdentityTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            identity2.ResourceId = resourceId2;
+            Assert.AreEqual(-1, identity1.CompareTo(identity2));
+        }
+
+        [TestCase]
+        public void EqualsMethodTrue()
+        {
+            Identity identity1 = new Identity();
+            Identity identity2 = new Identity();
+
```
markcowl, line 59: See comment below on equivalence classes, which would apply to equivalence and non-equivalence

azure-proto-core/Resources/Identity.cs
```diff
@@ -1,11 +1,9 @@
...
-using System.Diagnostics.CodeAnalysis;
 
 namespace azure_proto_core
 {
     /// <summary>
     /// Represents a managed identity
-    /// TODO: fill in properties, implement comparison and equality methods and operator overloads
     /// </summary>
     public class Identity : IEquatable<Identity>, IComparable<Identity>
     {
```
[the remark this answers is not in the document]
markcowl: We should be able to cover this with VirtualMachine, which should have a property of type Identity

markcowl (changes requested):

nisha-bhatia:
I have removed the test files from this PR, will submit a different PR for IdentityTests, SystemAssignedTests, and UserAssignedTests.

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,91 @@
...
+        public SystemAssignedIdentity(Guid tenantId, Guid principalId)
+        {
+            this.TenantId = tenantId;
+            this.PrincipalId = principalId;
+        }
+
+        public int CompareTo(SystemAssignedIdentity other)
+        {
+            if ((this.TenantId.HasValue == false && this.PrincipalId.HasValue == false) && other == null)
+                return 0;
```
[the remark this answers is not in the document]
markcowl:
``` c#
private static int NullableGuidCompare(Guid? first, Guid? second)
{
  if (!first.HasValue)
  {
     // null < any value
     return second.HasValue? -1 : 0;
  }
  if (!second.HasValue)
  {
    return 1;
  }

  return first.CompareTo(second);
}
```

Then you can use this in the comparison as follows:

``` c#

if (other == null)
{
   return null;
}
var tenantCompare = NullableGuidCompare(TenantId, other.TenantId);
if (tenantCompare != 0)
{
  return tenantCompare;
}

return NullableGuidCompare(PrincipalId, other.PrincipalId)
```

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,148 @@
...
+
+        public UserAssignedIdentity (Guid clientId, Guid principalId)
+        {
+            this.ClientId = clientId;
+            this.PrincipalId = principalId;
+        }
+
+        public int CompareTo (UserAssignedIdentity other)
+        {
+            if ((this.ClientId == Guid.Empty && this.PrincipalId == Guid.Empty) && other == null)
```
markcowl, line 23: see comments above

azure-proto-core/Shared/Autorest/CodeGenTypeAttribute.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+
+using System;
+
+namespace Azure.Core
+{
+    [AttributeUsage(AttributeTargets.Class)]
```
markcowl, line 10: If this is copied from Azure.Core, it may not have the intended effect - the type identity for this type will be different, meaning any code that have that looks for this attribute will not find it.

azure-proto-core/Shared/Autorest/CodeGenSuppressAttribute.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Enum | AttributeTargets.Struct, AllowMultiple = true)]
+    internal class CodeGenSuppressAttribute : Attribute
```
markcowl, line 9: see comment below - if these are copied from Azure.Core, they may not work

azure-proto-core/Shared/Autorest/CodeGenModelAttribute.cs
```diff
@@ -0,0 +1,31 @@
...
+// Licensed under the MIT License.
+
+#nullable enable
+
+using System;
+
+namespace Azure.Core
+{
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Enum | AttributeTargets.Struct)]
+    internal class CodeGenModelAttribute : CodeGenTypeAttribute
```
markcowl, line 11: see comments on attributes below

azure-proto-core/Shared/Autorest/CodeGenMemberAttribute.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#nullable enable
+
+using System;
+
+namespace Azure.Core
+{
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field)]
```
markcowl, line 10: see comments on attributes below

azure-proto-core/Shared/Autorest/CodeGenClientAttribute.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+
+namespace Azure.Core
+{
+    [AttributeUsage(AttributeTargets.Class)]
```
markcowl, line 8: see comments on attributes below

azure-proto-core/Shared/Autorest/PageableHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Azure.Core
+{
+    internal static class PageableHelpers
+    {
+        public static Pageable<T> CreateEnumerable<T>(Func<int?, Page<T>> firstPageFunc, Func<string?, int?, Page<T>>? nextPageFunc, int? pageSize = default) where T : notnull
```
markcowl, line 14: We have duplicated much of this code in our own types, are these necessary?

azure-proto-core/Shared/Autorest/PageableHelpers.cs
```diff
@@ -0,0 +1,97 @@
...
+                {
+                    Page<T> pageResponse = await pageFunc(continuationToken, pageSize).ConfigureAwait(false);
+                    yield return pageResponse;
+                    continuationToken = pageResponse.ContinuationToken;
+                    pageFunc = _nextPageFunc;
+                } while (!string.IsNullOrEmpty(continuationToken) && pageFunc != null);
+            }
+        }
+
+        internal class FuncPageable<T> : Pageable<T> where T : notnull
```
markcowl, line 64: same as above - much of this has already been moved into our own types in the core library

azure-proto-sdk.sln
```diff
@@ -13,7 +13,9 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "azure-proto-network", "azur
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "LegacyClient", "LegacyClient\LegacyClient.csproj", "{2361ABBC-0526-4920-A7F9-08E60BE6AF36}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "azure-proto-core-test", "azure-proto-core-test\azure-proto-core-test.csproj", "{5F200CAA-8A98-4A47-AD21-750C716E80DE}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "azure-proto-core-test", "azure-proto-core-test\azure-proto-core-test.csproj", "{5F200CAA-8A98-4A47-AD21-750C716E80DE}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Azure.Core", "..\..\Azure\azure-sdk-for-net\sdk\core\Azure.Core\src\Azure.Core.csproj", "{23207843-C2C4-4BEB-BE57-B21662E5CA9D}"
```
markcowl, line 18: we should not add this project to the solution

client/Scenarios/CreateVMForIdentity.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public override void Execute()
+        {
+            var client = new ArmClient();
+            var subscription = client.Subscription(Context.SubscriptionId);
+
+            // Create Resource Group
+            Console.WriteLine($"--------Start create group IDENTITY TEST {Context.RgName}--------");
+
+            // Create Resource Group
```
markcowl, line 22: any reason we need to propagate this commented code?

markcowl (changes requested):

azure-proto-core/Resources/Identity.cs
```diff
@@ -1,32 +1,145 @@
...
 {
     /// <summary>
     /// Represents a managed identity
-    /// TODO: fill in properties, implement comparison and equality methods and operator overloads
     /// </summary>
-    public class Identity : IEquatable<Identity>, IComparable<Identity>
+    public class Identity : IEquatable<Identity>, IComparable<Identity>, IUtf8JsonSerializable
     {
-        public Guid TenantId { get; set; }
+        private const string SystemAssigned = "SystemAssigned";
```
allenjzhang, line 19: Let's use flags enum for these.

azure-proto-core/Resources/Identity.cs
```diff
@@ -1,32 +1,145 @@
...
+                if (userCompare == 1 && (systemCompare == 0 || systemCompare == -1))
+                    return 1;
+                else if (userCompare == 0 && systemCompare == 0)
+                    return 0;
+                else
+                    return -1;
+            }
         }
 
         public bool Equals(Identity other)
```
allenjzhang, line 66: Please take a look at the code sample I sent you.

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            JsonElement rootElement = document.RootElement;
+            var identityJsonProperty = rootElement.EnumerateObject().First<JsonProperty>();
+            Assert.Throws<InvalidOperationException>(delegate { SystemAssignedIdentity.Deserialize(identityJsonProperty.Value); });
+        }
+
+        [TestCase]
+        public void TestDeserializerInvalid()
+        {
+            string json = "";
+            string path = ((AssemblyMetadataAttribute)GetType().Assembly.GetCustomAttribute(typeof(AssemblyMetadataAttribute))).Value;
```
bquantump, line 208: Set this in OneTimeSetup

azure-proto-core-test/azure-proto-core-test.csproj
```diff
@@ -18,12 +18,17 @@
...
+      <CoverageFiles Include="coverage.$(TargetFrameworks).cobertura.xml" />
     </ItemGroup>
-    <Exec Command= "dotnet $(UserProfile)\.nuget\packages\reportgenerator\4.7.1\tools\$(TargetFrameworks)\ReportGenerator.dll -reports:@(CoverageFiles) -targetdir:coverage -reporttypes:Html"/>
+    <Exec Command="dotnet $(UserProfile)\.nuget\packages\reportgenerator\4.7.1\tools\$(TargetFrameworks)\ReportGenerator.dll -reports:@(CoverageFiles) -targetdir:coverage -reporttypes:Html" />
   </Target>
   <ItemGroup>
     <ProjectReference Include="..\azure-proto-core\azure-proto-core.csproj" />
   </ItemGroup>
-
+  <ItemGroup Condition="'$(IsTestProject)' == 'true'">
```
bquantump, line 28: This is a test project, so why check?

nisha-bhatia:
Screenshot 2020-12-10 173300

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,348 @@
...
+using System.Text.Json;
+
+namespace azure_proto_core_test
+{
+    public class SystemAssignedIdentityTests : SystemAssignedIdentity
+    {
+
+        [TestCase(null, null, null, null, true)]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", false)]
+        public void CompareToZero(string tenantId1, string principalId1, string tenantId2, string principalId2, bool isNull)
```
m-nash, line 16: don't think  we need the extra isNull parameters, we can check the tenantId1 param itself.  This allows us to compare null to not null

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,348 @@
...
+            var identityJsonProperty = rootElement.EnumerateObject().First();
+            Assert.Throws<InvalidOperationException>(delegate { Deserialize(identityJsonProperty.Value); });
+        }
+
+        [TestCase]
+        public void TestDeserializerTenantIdInvalid()
+        {
+            string json = "";
+            string path = ((AssemblyMetadataAttribute)GetType().Assembly.GetCustomAttribute(typeof(AssemblyMetadataAttribute))).Value;
+            using (StreamReader f = new StreamReader(path + "/TestAssets/SystemAssignedIdentity/SystemAssignedInvalid.json"))
```
m-nash, line 254: pretty sure you can use "./TestAssets/SystemAssignedIdentity/SystemAssignedInvalid.json" since the file will be copied local along side the binaries.

if this isn't happening then you need to set the content property of the json file

Also there is a simplified version of this
var json = File.ReadAllText("./TestAssets/SystemAssignedIdentity/SystemAssignedInvalid.json");

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,348 @@
...
+            using (StreamReader f = new StreamReader(path + "/TestAssets/SystemAssignedIdentity/SystemAssignedOneEmptyString.json"))
+                json = f.ReadToEnd();
+            JsonDocument document = JsonDocument.Parse(json);
+            JsonElement rootElement = document.RootElement;
+            var identityJsonProperty = rootElement.EnumerateObject().First();
+            Assert.Throws<FormatException>(delegate { Deserialize(identityJsonProperty.Value); });
+        }
+
+        [TestCase]
+        public void TestDeserializerTenantIdValueNull()
```
m-nash, line 224: these can be parameterized by passing in the path string since everything else appears the same

azure-proto-core-test/azure-proto-core-test.csproj
```diff
@@ -20,10 +20,15 @@
...
       <CoverageFiles Include="coverage.$(TargetFrameworks).cobertura.xml"/>
     </ItemGroup>
-    <Exec Command= "dotnet $(UserProfile)\.nuget\packages\reportgenerator\4.7.1\tools\$(TargetFrameworks)\ReportGenerator.dll -reports:@(CoverageFiles) -targetdir:coverage -reporttypes:Html"/>
+    <Exec Command=" dotnet $(UserProfile)\.nuget\packages\reportgenerator\4.7.1\tools\$(TargetFrameworks)\ReportGenerator.dll -reports:@(CoverageFiles) -targetdir:coverage -reporttypes:Html" />
   </Target>
   <ItemGroup>
     <ProjectReference Include="..\azure-proto-core\azure-proto-core.csproj" />
   </ItemGroup>
-
+  <ItemGroup>
```
m-nash, line 28: not needed with setting the properties on the json files

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,155 @@
...
+            writer.WriteEndObject();
+            writer.Flush();
+        }
+
+        public static bool Equals(SystemAssignedIdentity original, SystemAssignedIdentity other)
+        {
+            if (original == null)
+                return other == null;
+
+            if (other == null)
```
m-nash, line 146: line 146 to 150 are redundant they will be executed inside original.Equals(other)

azure-proto-core/azure-proto-core.csproj
```diff
@@ -9,6 +9,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Azure.Core" Version="1.6.0" />
```
m-nash, line 12: why was this reference needed?

m-nash (changes requested):

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,155 @@
...
+            else if ((isPrincipalIdNull && !isTenantIdNull) || (!isPrincipalIdNull && isTenantIdNull))
+                throw new InvalidOperationException("Either TenantId or PrincipalId were null");
+
+            else
+                return new SystemAssignedIdentity(tenantId, principalId);
+        }
+
+        public static void Serialize(Utf8JsonWriter writer, SystemAssignedIdentity systemAssignedIdentity)
+        {
+            writer.WriteStartObject();
```
m-nash, line 122: parameter checking for null

m-nash (changes requested):

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            return -1;
+        }
+
+        public bool Equals(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return false;
+
+            if (TenantId.HasValue == false &&
```
bquantump, line 54: public Guid? TenantId { get; set; }

      public Guid? PrincipalId { get; set; }

These are nullable and you aren't checking for null here.

The the empty constructor has these nulled.

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,144 @@
...
+                PrincipalId.HasValue == false &&
+                other.TenantId.HasValue == false &&
+                other.PrincipalId.HasValue == false)
+            {
+                return true;
+            }
+
+            if (!TenantId.HasValue && !PrincipalId.HasValue)
+                return false;
```
bquantump, line 63: What if TenatId has value but PrincipleID does not? 

Equals on line 68 would fail no?

bquantump: Then no need to check both right?

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,131 @@
...
+            PrincipalId = principalId;
+        }
+
+        public int CompareTo(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return 1;
+
+            if (TenantId.HasValue == false && other.TenantId.HasValue == false)
+                return 0;
```
bquantump, line 27: Use ! instead of false like bellow to be consistent

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            return -1;
+        }
+
+        public bool Equals(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return false;
+
+            if (TenantId.HasValue == false && other.TenantId.HasValue == false)
```
bquantump, line 49: Use ! instead of false like bellow to be consistent

bquantump (approved):

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+using System.Text.Json;
+
+namespace azure_proto_core_test
+{
+    public class SystemAssignedIdentityTests : SystemAssignedIdentity
+    {
+        [TestCase(0, null, null, null, null)]
+        [TestCase(0, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98")]
+        [TestCase(1, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", null, null)]
+        [TestCase(1, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db46", "de29bab1-49e1-4705-819b-4dfddceaaa97")]
```
markcowl, line 15: I think it would be useful to have test cases in which both PrincipalId and TenantId are non-null, and:
- Tid1 == Tid2, Pid1 < P id2 (result -1)
- Tid1 == Tid2, Pid1 > Pid2 (result 1)
- Tid1 < Tid2, Pid1 == Pid2 (result -1)
- Tid1 > Tid1, Pid1 == Pid2 (result 1)
- Tid1 < Tid2, Pid1 > Pid2 (result -1)
- Tid1 > Tid2, Pid1 < Pid2 (result 1)

Essentially, think of completing a table like:


| CompareTo(Tid1, Tid2) | CompareTo(Pid1, Pid2 | Result |
| ---: | ---------------------------------: | -----: |
| -1 | -1 | -1 |
| -1 |  0 | -1 |
| -1 | 1 | -1|
| 0 | -1 | -1 |
| 0 | 0 | 0 |
| 0 | 1 | 1 |
| 1 | -1 | 1 |
| 1 | 0 | 1 |
|1 | 1 | 1 |

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        [TestCase(-1, null, null, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98")]
+        [TestCase(-1, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db48", "de29bab1-49e1-4705-819b-4dfddceaaa99")]
+        public void CompareTo(int answer, string tenantId1, string principalId1, string tenantId2, string principalId2)
+        {
+            SystemAssignedIdentity identity1;
+            SystemAssignedIdentity identity2;
+            if (tenantId1 == null)
+                identity1 = new SystemAssignedIdentity();
```
markcowl, line 23: Use braces in the body of conditionals, especially if there is an if/else condition

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+
+        [TestCase(null, null, null, null)]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98")]
+        public void EqualsMethodTrue(string tenantId1, string principalId1, string tenantId2, string principalId2)
+        {
+            SystemAssignedIdentity identity1;
+            SystemAssignedIdentity identity2;
+            if (tenantId1 == null)
+                identity1 = new SystemAssignedIdentity();
```
markcowl, line 44: same comment

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+            else
+                identity2 = new SystemAssignedIdentity(new Guid(tenantId2), new Guid(principalId2));
+
+            Assert.IsTrue(identity1.Equals(identity2));
+        }
+
+        [TestCase(null, null, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98")]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db44", "de29bab1-49e1-4705-819b-4dfddceaaa94")]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", null, null)]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa93", "72f988bf-86f1-41af-91ab-2d7cd011db42", "de29bab1-49e1-4705-819b-4dfddceaaa91")]
```
markcowl, line 61: same comment as above.  It would be good to have a complete representation of the equivalence classes:
- Tid1 == Tid2, Pid1 != Pid2
- Tid != Tid2, Pid1 == Pid2
- Tid1 != Tid2, Pid1 != Pid2

In the above, you have good coverage of the possible nulls, but have two examples of both values being unequal, rather than mixing one element equal and one element unequal variations

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        [TestCase]
+        public void TestDeserializerTenantIdInvalid()
+        {
+            var identityJsonProperty = DeserializerHelper("SystemAssignedInvalid.json");
+            Assert.Throws<InvalidOperationException>(delegate { Deserialize(identityJsonProperty.Value); });
+        }
+
+        [TestCase]
+        public void TestSerializer()
```
markcowl, line 193: SHould use a more descriptive name 'TestSerializeValidIdentity

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
+﻿using azure_proto_core.Resources;
+using NUnit.Framework;
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+namespace azure_proto_core_test
+{
+    public class SystemAssignedIdentityTests : SystemAssignedIdentity
```
markcowl, line 10: Generally, having your test class extend the class that is the subject of the test is a pattern to avoid.  For one thing, it prevents your test from extending any test class that might be required for your framework.  For another, it can be confusing which methods are test methods, and which are inherited methods.

Ifif you do need a test class the extends SystemAssignedIdentity, it is generally better to have a separate 'TestSystemAssignedIdentity' class that is used by this class.

azure-proto-core-test/SystemAssignedIdentityTests.cs
```diff
@@ -0,0 +1,299 @@
...
+        {
+            SystemAssignedIdentity identity1 = null;
+            SystemAssignedIdentity identity2 = new SystemAssignedIdentity();
+            Assert.IsFalse(Equals(identity1, identity2));
+        }
+
+        [TestCase]
+        public void TestEqualsOtherIdentityNull()
+        {
+            SystemAssignedIdentity identity1 = new SystemAssignedIdentity();
```
markcowl, line 270: How do these tests differ from the equality tests above?  What is being tested in addition?

markcowl (changes requested):

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            if (principalId == default(Guid) || tenantId == default(Guid))
+                throw new InvalidOperationException("Either TenantId or PrincipalId were null");
+
+            return new SystemAssignedIdentity(tenantId, principalId);
+        }
+
+        public static void Serialize(Utf8JsonWriter writer, SystemAssignedIdentity systemAssignedIdentity)
+        {
+            if (systemAssignedIdentity == null || writer == null)
```
m-nash, line 95: should have two different checks with two different argumentnullexceptions

The code should look like this
```suggestion
            if (systemAssignedIdentity == null)
                throw new ArgumentNullException(nameof(systemAssignedIdentity));
```
This keeps the exception up to date if the name of the parameter ever changes.

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,125 @@
...
+
+            return -1;
+        }
+
+        public bool Equals(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return false;
+
+            if (!TenantId.HasValue && !other.TenantId.HasValue)
```
m-nash, line 49: Can we simplify to
```suggestion
           return TenantId.Equals(other.TenantId) && PrincipalId.Equals(other.PrincipalId);
```

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,125 @@
...
+            TenantId = tenantId;
+            PrincipalId = principalId;
+        }
+
+        public int CompareTo(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return 1;
+
+            if (!TenantId.HasValue && !other.TenantId.HasValue)
```
m-nash, line 26: Can we simplify to
```suggestion
            int compareResult = 0;
            if ((compareResult = TenantId.GetValueOrDefault().CompareTo(other.TenantId.GetValueOrDefault())) == 0 &&
                (compareResult = PrincipalId.GetValueOrDefault().Compare(other.PrincipalId.GetValueOrDefault())) == 0)
            {
                return 0;
            }

            return compareResult;
```

m-nash (changes requested):

nisha-bhatia:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4467

m-nash (approved):

azure-proto-core/Resources/SystemAssignedIdentity.cs
```diff
@@ -0,0 +1,116 @@
...
+            PrincipalId = principalId;
+        }
+
+        public int CompareTo(SystemAssignedIdentity other)
+        {
+            if (other == null)
+                return 1;
+
+            int compareResult = 0;
+            if ((compareResult = PrincipalId.GetValueOrDefault().CompareTo(other.PrincipalId.GetValueOrDefault())) == 0 &&
```
markcowl, line 27: if `this.PrincipalId != other.PrincipalId`, this will return the value of `this.PrincipalId.CompareTo(other.PrincipalId)`, due to short circuiting.  Is this the intention? (That is, the algorithm first compares principalId and only compares TenantId if PrincipalIds are equal.  If we think about using this for sorting, I think, sense Tenant is the container for ResourceId, this should be the other way around (that is Principals from the same Tenant should sort together).

markcowl (changes requested):

markcowl (approved):

nisha-bhatia:
Updated ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4999

PR #47, opened 2020-11-17
Remove need for registry and use activator instead

m-nash:


azure-proto-core-test/ResourceGroupOperationsTests.cs
```diff
@@ -0,0 +1,23 @@
...
+
+namespace azure_proto_core_test
+{
+    public class ResourceGroupOperationsTests
+    {
+        [TestCase]
+        public void CreateResourceFromModel()
+        {
+            //TODO: 4622 needs complete with a Mocked example to fill in this test
+            //public ArmResponse<TOperations> CreateResource<TContainer, TOperations, TResource>(string name, TResource model, azure_proto_core.Location location = default)
```
bquantump, line 11: Why not call CreateResource, CreateResourceFromModel? I had to read the code a bit to figure out that this is the use case for these methods.

m-nash: Lets take this in a separate issue since this PR didn't introduce the naming of those libraries.  There will be an in depth naming review during arch board.

azure-proto-core/ArmClient.cs
```diff
@@ -202,11 +184,7 @@ public ResourceGroupOperations ResourceGroup(PhResourceGroup resourceGroup)
                 location = azure_proto_core.Location.Default;
             }
 
-            TContainer container;
-            if (!Registry.TryGetContainer<TContainer, ArmResource, TOperations, TResource>(this.ClientContext, new ArmResource($"/subscriptions/{subscription}/resourceGroups/{resourceGroup}", location), out container))
-            {
-                throw new InvalidOperationException($"No resource type matching '{typeof(TResource)}' found.");
-            }
+            TContainer container = Activator.CreateInstance(typeof(TContainer), ClientContext, new ArmResource($"/subscriptions/{subscription}/resourceGroups/{resourceGroup}", location)) as TContainer;
```
weshaggard, line 187: This seems reasonable and better then having a manual registry. The only reason a manual registry might be interesting is if using reflection becomes a bottleneck but I don't think that will be the case for these. 

This is assuming that all the types you are creating has a particular constructor signature, do we have any guarantees around that? If not we might want to had some error handling here.

weshaggard (approved):
Looks reasonable. Added a comment about potential error handling.

azure-proto-core/ArmClient.cs
m-nash: We do have guarantees around that all of them will be auto-generated

PR #48, opened 2020-11-17
New class hier for get and delete operations

bquantump:


azure-proto-core/Adapters/ModelAdapters.cs
```diff
@@ -24,4 +24,22 @@ protected TrackedResource(ResourceIdentifier id, Location location, T data)
             return other.Model;
         }
     }
+
+    //Or call generic resource, other resource??
+    public abstract class ChildResource<T> : Resource where T : class
```
markcowl, line 29: I think we should go with SubResource - this is the term used most often in Swagger

azure-proto-core/Adapters/ModelAdapters.cs
```diff
@@ -24,4 +24,22 @@ protected TrackedResource(ResourceIdentifier id, Location location, T data)
             return other.Model;
         }
     }
+
+    //Or call generic resource, other resource??
```
markcowl, line 28: For each of these classes with Type arguments, we should consider having a base class without type arguments (i.e. ChildResource)  This allows us to write code to address all SubResources as a class (for example, in method signatures).  This is a pattern often followed in the .net framework (for example, IList, IList<T>).

azure-proto-core/DeletableResourceOperations.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    /// <summary>
+    /// Implementation for resources that implement the immutable resource pattern
+    /// </summary>
+    public abstract class DeletableResourceOperations<TResource, TOperations> : GenericResourcesOperations <TResource, TOperations>
```
markcowl, line 13: We should be consistent in naming (GenericResourceOperations, DeletableResourceOperations) or (GenericResourcesOperations, DeletableResourcesOperations)

azure-proto-core/GenericResourcesOperations.cs
```diff
@@ -0,0 +1,52 @@
...
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    /// <summary>
+    /// Implementation for resources that implement the immutable resource pattern
+    /// </summary>
+    public abstract class GenericResourcesOperations<TResource, TOperations> : OperationsBase
```
markcowl, line 13: Can this just be ResourceOperations?

azure-proto-core/DeletableResourceOperations.cs
markcowl: Just a thought - alternately, what would it look like to have interfaces for Create and Delete operations, rather than a class hierarchy?  In this case, we would have ResourcesOperations, and separate interfaces for Create, Delete, and Tagging operations?  The downside is that any change in the operations would be a breaking change, but the advantage would be that they can be applied in any pattern to a resource.

markcowl (changes requested):

azure-proto-core/ResourceOperationsBase.cs
```diff
@@ -9,9 +9,9 @@ namespace azure_proto_core
     /// TODO: Split into ResourceOperations/TrackedResourceOperations
     /// </summary>
     /// <typeparam name="Model"></typeparam>
-    public abstract class ResourceOperationsBase<TOperations, TResource> : OperationsBase
-        where TResource : Resource
-        where TOperations : ResourceOperationsBase<TOperations, TResource>
+    public abstract class ResourceOperationsBase<TOperations, TResource> : DeletableResourceOperations <TResource, TOperations>
```
m-nash, line 12: Isn't this a TrackedResource at this point?

azure-proto-core/GenericResourcesOperations.cs
m-nash: yeah Generic has other meaning when it comes to results from Arm vs results from an RP

azure-proto-core/ResourceOperationsBase.cs
```diff
@@ -9,9 +9,9 @@ namespace azure_proto_core
...
+        where TResource:Resource 
+        where TOperations: DeletableResourceOperations<TResource, TOperations>
```
m-nash, line 14: Does this mean that a resource that has tags has to be deletable?  Is that always the case?

azure-proto-core/DeletableResourceOperations.cs
m-nash: > We should be consistent in naming (GenericResourceOperations, DeletableResourceOperations) or (GenericResourcesOperations, DeletableResourcesOperations)

I suggest using Resource Singular for both

m-nash (changes requested):

PR #50, opened 2020-11-19
Implement comparison methods in Sku class and add its tests

YalinLi0312:
The Sku class properties comparison priority I followed: Name > Family > Size > Tier > Capacity

Code coverage result in VisualStudio:
Screenshot 2020-11-18 154916

Code coverage result in Cobertura:
Screenshot 2020-11-18 155304

### DIFF
--- a/azure-proto-core-test/SkuTests.cs
+++ b/azure-proto-core-test/SkuTests.cs
@@ -1,0 +1,135 @@
+﻿using azure_proto_core;
+using NUnit.Framework;
+
+namespace azure_proto_core_test
+{
+    class SkuTests
+    {
+        [Test]
+        public void CompareToTests()
+        {
+            Sku sku1 = new Sku();
+            Sku sku2 = null;
+
+            sku1.Name = "name";
+            sku1.Family = "family";
+            sku1.Size = "size";
+            sku1.Tier = "tier";
+            Assert.AreEqual(1, sku1.CompareTo(sku2));
+
+            sku2 = new Sku();
+            sku2.Name = "name";
+            sku2.Family = "family";
+            sku2.Size = "size";
+            sku2.Tier = "tier";
+            Assert.AreEqual(0, sku1.CompareTo(sku2));
+
+            sku1.Capacity = 1;
+            Assert.AreEqual(1, sku1.CompareTo(sku2));
+
+            sku2.Capacity = 2;
+            Assert.AreEqual(-1, sku1.CompareTo(sku2));
+
+            sku2.Capacity = 0;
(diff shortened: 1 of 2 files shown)


azure-proto-core-test/SkuTests.cs
```diff
@@ -0,0 +1,135 @@
+﻿using azure_proto_core;
+using NUnit.Framework;
+
+namespace azure_proto_core_test
+{
+    class SkuTests
+    {
+        [Test]
+        public void CompareToTests()
```
m-nash, line 9: Break these out into tests for each parameter and use parameterized testing like we discussed in our sync today

azure-proto-core-test/SkuTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            Sku sku1 = new Sku();
+            Sku sku2 = null;
+
+            sku1.Name = "name";
```
m-nash, line 14: Verify with @markcowl where the constraints are documented for this object?  It is possible this is deserialize only in which case that is "irrelevant"?

m-nash (changes requested):

azure-proto-core-test/SkuTests.cs
YalinLi0312: Updated

azure-proto-core-test/SkuTests.cs
YalinLi0312: Pending until getting the exact constraints from the contact. Will update in a separate PR if need to.

YalinLi0312:
New code coverage results:
Screenshot 2020-11-20 121344
Screenshot 2020-11-20 121426

YalinLi0312:
Update code coverage results:
Screenshot 2020-11-20 163215
Screenshot 2020-11-20 163259

azure-proto-core/Resources/Sku.cs
```diff
@@ -17,12 +16,30 @@ public class Sku : IEquatable<Sku>, IComparable<Sku>
 
         public int CompareTo(Sku other)
         {
-            throw new NotImplementedException();
+            if (other == null) return 1;
+            if (this.Name == null && other.Name != null) return -1;
+            if (this.Name != null && this.Name.CompareTo(other.Name) != 0) return this.Name.CompareTo(other.Name);
```
m-nash, line 21: store this.Name.CompareTo(other.Name) so you don't have to call it twice

this applies to all the other items in this method and Equals

m-nash (changes requested):

bquantump (approved):

PR #51, opened 2020-11-19
Implement comparison methods in Plan class and add its tests

YalinLi0312:
The Plan class properties comparison priority I followed: Name > Product > PromotionCode > Publisher > Version

Code coverage result in VisualStudio:
Screenshot 2020-11-18 154815

Code coverage result in Cobertura:
Screenshot 2020-11-18 161211

### DIFF
--- a/azure-proto-core/Resources/Plan.cs
+++ b/azure-proto-core/Resources/Plan.cs
@@ -18,15 +18,35 @@ namespace azure_proto_core
         {
             if (other == null) return 1;
             if (this.Name == null && other.Name != null) return -1;
-            if (this.Name != null && this.Name.CompareTo(other.Name) != 0) return this.Name.CompareTo(other.Name);
+            if (this.Name != null)
+            {
+                int compareNameResult = this.Name.CompareTo(other.Name);
+                if (compareNameResult != 0) return compareNameResult;
+            }
             if (this.Product == null && other.Product != null) return -1;
-            if (this.Product != null && this.Product.CompareTo(other.Product) != 0) return this.Product.CompareTo(other.Product);
+            if (this.Product != null)
+            {
+                int compareProductResult = this.Product.CompareTo(other.Product);
+                if (compareProductResult != 0) return compareProductResult;
+            }
             if (this.PromotionCode == null && other.PromotionCode != null) return -1;
-            if (this.PromotionCode != null && this.PromotionCode.CompareTo(other.PromotionCode) != 0) return this.PromotionCode.CompareTo(other.PromotionCode);
+            if (this.PromotionCode != null)
+            {
+                int comparePromotionCodeResult = this.PromotionCode.CompareTo(other.PromotionCode);
+                if (comparePromotionCodeResult != 0) return comparePromotionCodeResult;
+            }
             if (this.Publisher == null && other.Publisher != null) return -1;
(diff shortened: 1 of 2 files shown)


azure-proto-core/Resources/Plan.cs
```diff
@@ -17,12 +16,43 @@ public class Plan : IEquatable<Plan>, IComparable<Plan>
 
         public int CompareTo(Plan other)
         {
-            throw new NotImplementedException();
+            if (other == null) return 1;
+            if (this.Name.CompareTo(other.Name) == 0)
+            {
```
bquantump, line 21: This nesting is pretty intense and hard to read. This can be accomplished without this nesting. 

if not x(){
Return xxx
}
if not y(){
return yyy
}

last line would then be the nested line:   return this.Version.CompareTo(other.Version);

azure-proto-core/Resources/Plan.cs
```diff
@@ -17,12 +16,43 @@ public class Plan : IEquatable<Plan>, IComparable<Plan>
...
+            }
+            return this.Name.CompareTo(other.Name);
         }
 
         public bool Equals(Plan other)
         {
-            throw new NotImplementedException();
+            if (other == null) return false;
+            if (this.Name.Equals(other.Name))
+            {
```
bquantump, line 43: Why the super nested if checks here? Can add these more into the one line.

YalinLi0312:
New code coverage results:
Screenshot 2020-11-20 150131
Screenshot 2020-11-20 150136

azure-proto-core/Resources/Plan.cs
YalinLi0312: Here comparing in this sequence: Name > Product > PromotionCode > Publisher > Version. For example, when name's value is not same, will skip other properties comparisons.

bquantump: && has the same effect, no?

azure-proto-core/Resources/Plan.cs
YalinLi0312: Good point, will update

azure-proto-core/Resources/Plan.cs
YalinLi0312: If we don't handle the case when a property value is null, we can use &&. Now the comparison is more complex, so I have updated to without using nesting if statements.

YalinLi0312:
Update code coverage results:
Screenshot 2020-11-20 155840
Screenshot 2020-11-20 155926

azure-proto-core/Resources/Plan.cs
```diff
@@ -17,12 +16,29 @@ public class Plan : IEquatable<Plan>, IComparable<Plan>
 
         public int CompareTo(Plan other)
         {
-            throw new NotImplementedException();
+            if (other == null) return 1;
+            if (this.Name == null && other.Name != null) return -1;
+            if (this.Name != null && this.Name.CompareTo(other.Name) != 0) return this.Name.CompareTo(other.Name);
```
m-nash, line 21: Same comment from Sku PR.  Store result of this.Name.CompareTo(other.Name) so you don't have to call it twice

m-nash (changes requested):

bquantump (approved):

PR #52, opened 2020-11-19
Interface prototype changes (Includes subnet update as well)

bquantump:


azure-proto-core/Adapters/ModelAdapters.cs
```diff
@@ -24,4 +24,22 @@ protected TrackedResource(ResourceIdentifier id, Location location, T data)
             return other.Model;
         }
     }
+
+    //Or call generic resource, other resource??
+    public abstract class ChildResource<T> : Resource where T : class
```
markcowl, line 29: I think we should use SubResource or ProxyResource instead of ChildResource

azure-proto-core/DeletableResourceOperations.cs
```diff
@@ -0,0 +1,17 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    public interface IDeletableResource<TOperations, TResource> 
+        where TResource:Resource 
+        where TOperations: IDeletableResource<TOperations, TResource>
+    {
+        ArmOperation<Response> Delete();
```
markcowl, line 14: Should we have StartDelete, StartDeleteAsync, Delete, DeleteAsync?

azure-proto-core/GenericResourcesOperations.cs
```diff
@@ -0,0 +1,49 @@
+using Azure;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    public abstract class GenericResourcesOperations<TOperations, TResource> : OperationsBase
```
markcowl, line 10: Why not just ResourcesOperations?  Also, I like ResourceOperations better, unless there is a naming rule that prevents this.

azure-proto-core/TagableResourceOperations.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Azure;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    public interface ITagable<TOperations, TResource>
```
markcowl, line 7: ITaggable

markcowl (changes requested):

azure-proto-core/DeletableResourceOperations.cs
bquantump: Doesn't Startxxx only apply to certain RPs that do LROs?

azure-proto-core/ResourceOperations.cs
```diff
@@ -0,0 +1,49 @@
+using Azure;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace azure_proto_core
+{
+    public abstract class ResourceOperations<TOperations, TResource> : OperationsBase
```
m-nash, line 10: rename back to ResourceOperationsBase

azure-proto-core/Adapters/ModelAdapters.cs
m-nash: SubResource makes more sense

m-nash (changes requested):

azure-proto-core/DeletableResourceOperations.cs
bquantump: I think we should do this a seperate task since, VirtualNetworkOperations, NetworkInterfaceOperations, ArmOperation,  and VirtualMachineOperations all need update to implement this interface when we add StartDelete.

bquantump: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_sprints/backlog/DotNET%20Management%20SDK%20Team/DotNET%20Management%20SDK/FE/180?workitem=4883

m-nash (approved):

PR #54, opened 2020-11-25
Subscription Container operations (fix get model if newer too)

bquantump:


azure-proto-core/SubscriptionContainerOperations.cs
```diff
@@ -0,0 +1,46 @@
...
+namespace azure_proto_core
+{
+    /// <summary>
+    /// Subscription Container Operations
+    /// </summary>
+    public class SubscriptionContainerOperations : OperationsBase
+    {
+        public static readonly string AzureResourceType = "Microsoft.Resources/subscriptions";
+
+        public SubscriptionContainerOperations(ArmClientContext context, string defaultSubscription) : base(context, $"/subscriptions/{defaultSubscription}") { }
```
m-nash, line 17: typically the container takes the resource/id of its parent.  The parent of a subscription is loosely a tenant, but currently its armClient so I wonder if the only parameter to this class should be ArmClientContext?

azure-proto-core/SubscriptionContainerOperations.cs
```diff
@@ -0,0 +1,46 @@
...
+
+        public SubscriptionContainerOperations(ArmClientContext context, ResourceIdentifier id) : base(context, id) { }
+
+        public SubscriptionContainerOperations(ArmClientContext context, Resource subscription) : base(context, subscription) { }
+
+        public override ResourceType ResourceType => AzureResourceType;
+
+        public Pageable<SubscriptionOperations> List(CancellationToken cancellationToken = default)
```
m-nash, line 25: Didn't these exist as methods in armclient?  Should they be removed since list now lives here?

m-nash (changes requested):

azure-proto-core/SubscriptionContainerOperations.cs
bquantump: Yes they should be removed

m-nash (approved):

PR #62, opened 2020-12-08
add Autorest

nisha-bhatia:


bquantump:
Why are we adding the autorest dependency this way? Are we planning on make manual changes to the code?

nisha-bhatia:
Only adding this to get the serializer/deserializer working now, it will be deleted later.

PR #64, opened 2020-12-09
Simplify comparison methods implementations in Sku and Plan class

YalinLi0312:


azure-proto-core/Resources/Plan.cs
```diff
@@ -27,82 +27,17 @@ public int CompareTo(Plan other)
                 return 1;
             }
 
-            if (Name == null && other.Name != null)
+            int compareResult = 0;
+            if ((compareResult = string.Compare(Name, other.Name)) == 0 &&
```
allenjzhang, line 31: Please add StringComparison.InvariantCultureIgnoreCase as 3rd param to all string.Compare calls.

azure-proto-core/Resources/Plan.cs
```diff
@@ -112,34 +47,16 @@ public bool Equals(Plan other)
...
-
-            if (Publisher == null && other.Publisher != null ||
-                Publisher != null && !Publisher.Equals(other.Publisher))
-            {
-                return false;
-            }
-
-            if (Version == null && other.Version != null || Version != null && !Version.Equals(other.Version))
+            if (object.Equals(Name, other.Name) &&
+                object.Equals(Product, other.Product) &&
```
allenjzhang, line 51: here you don't need to compare individual properties. Just (this, other). This is supposed to be an optimization if this and other is same object, then true. For each properties, let it compare below.

azure-proto-core/Resources/Sku.cs
```diff
@@ -27,82 +27,16 @@ public int CompareTo(Sku other)
                 return 1;
             }
 
-            if (Name == null && other.Name != null)
+            int compareResult = 0;
+            if ((compareResult = string.Compare(Name, other.Name)) == 0 &&
```
allenjzhang, line 31: StringComparison.InvariantCultureIgnoreCase

azure-proto-core/Resources/Sku.cs
```diff
@@ -112,33 +46,16 @@ public bool Equals(Sku other)
...
             }
 
-            if (Name == null && other.Name != null || Name != null && !Name.Equals(other.Name))
-            {
-                return false;
-            }
-
-            if (Family == null && other.Family != null || Family != null && !Family.Equals(other.Family))
+            if (object.Equals(Name, other.Name) &&
+                object.Equals(Family, other.Family) &&
```
allenjzhang, line 50: same as in plan. Please use proper string compare for each property.

allenjzhang (changes requested):

azure-proto-core/Resources/Plan.cs
```diff
@@ -27,82 +27,17 @@ public int CompareTo(Plan other)
...
-                var compareVersionResult = Version.CompareTo(other.Version);
-
-                if (compareVersionResult != 0)
-                {
-                    return compareVersionResult;
-                }
-            }
-
-            return 0;
+            return compareResult;
```
m-nash, line 40: I believe this removes the ordering that we had in place.  Before Name had priority for comparison over Version as an example, so if i had 2 objects A and B and A.Name > B.Name but A.Version < B.Version compareTo will now return -1 instead of 1 since the last value in compareResult will be A.Version.CompareTo(B.Version).  This gets worse in the following scenario
A.Name > B.Name && A.Version == B.Version In this case again the expected value is 1, but the way the code is written the actual value will be 0.

Please add the example tests cases I have given here to prove the assertion either way.

azure-proto-core/Resources/Plan.cs
m-nash: object.Equals I believe does a ReferenceEquals only.  If there was a string.Equals(a, b) then that could be used here and avoid some of the null checking that was previously there

m-nash (changes requested):

azure-proto-core/Resources/Plan.cs
m-nash: A and B A.Name < B.Name && A.Version > B.Version expected -1

YalinLi0312: I tested the three scenarios, they all can get the expected value, The updates in this PR doesn't remove the ordering, it will compare Name first, if not equals will jump out the if statement and return `compareResult`, same as other properties.

azure-proto-core/Resources/Sku.cs
```diff
@@ -46,11 +51,16 @@ public bool Equals(Sku other)
                 return false;
             }
 
-            if (object.Equals(Name, other.Name) &&
-                object.Equals(Family, other.Family) &&
-                object.Equals(Size, other.Size) &&
-                object.Equals(Tier, other.Tier) &&
-                object.Equals(Capacity, other.Capacity))
+            if (object.Equals(this, other))
```
m-nash, line 54: Can we make this explicit since since thats all its really doing?  There are 4 instances of this.
```suggestion
            if (object.ReferenceEquals(this, other))
```

m-nash (changes requested):

m-nash (approved):

allenjzhang (approved):

PR #66, opened 2020-12-10
Unit test for ListAtContext

bquantump:


ClientTests/ResourceListOperationsTest.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace azure_proto_core_test
+{
+    public class ResourceListOperationsTest
+    {
+        [TestCase]
+        public void TestArmResonseArmResource()
+        {
+             TestListActivator<ArmResourceOperations, ArmResource>();
+        }
+        private static void TestListActivator<T, U>()
```
m-nash: Can we give the type parameters semantic names like TOperations etc?

ClientTests/ResourceListOperationsTest.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var testMethod = typeof(azure_proto_core.ResourceListOperations).GetMethod("CreateResourceConvertor", BindingFlags.Static | BindingFlags.NonPublic);
+            var asGeneric = testMethod.MakeGenericMethod(new System.Type[] {typeof(T), typeof(U)});
+            var context = new ArmClientContext(new Uri("https://management.azure.com"), new DefaultAzureCredential());
+            var function =  (Func<GenericResourceExpanded, T>) asGeneric.Invoke(null, new object[] {context, new ArmClientOptions()} );
+            var resource = new GenericResourceExpanded();
+            resource.Location = "East US";
+            resource.Tags =  new Dictionary<string, string>{ {"tag1", "value1"} };
+            function.DynamicInvoke(new object [] {resource});
```
m-nash: Can we trap the return type here which I believe should be a T and assert some of its properties.  Like it should have the same context and resource property values that we set on lines 23 and 25-27?

azure-proto-compute/Placeholder/AvailabilitySetData.cs
```diff
@@ -17,6 +17,8 @@ public AvailabilitySetData(Azure.ResourceManager.Compute.Models.AvailabilitySet
             }
         }
 
+
```
m-nash, line 20: I think you can revert this file

azure-proto-core/ResourceListOperations.cs
```diff
@@ -196,10 +191,17 @@ private static void Validate(ResourceIdentifier id)
         {
             return new PhWrappingAsyncPageable<GenericResourceExpanded, TOperations>(
                 result,
-                s => Activator.CreateInstance(
+                CreateResourceConvertor<TOperations, TResource>(clientContext, clientOptions));
+        }
+
+        private static Func<GenericResourceExpanded, TOperations> CreateResourceConvertor<TOperations, TResource>(ArmClientContext clientContext, ArmClientOptions clientOptions)
+        where TOperations : ResourceOperationsBase<TOperations, TResource>
```
m-nash, line 198: where clauses should be indented similar to the other methods

m-nash (changes requested):

markcowl (changes requested):
I think this is fine, modulo the spelling change, and I don't want to spend too much time on this, as we won't need this once we move to the .Net repo, but I wonder if this wouldn't be simpler if we simply had a constructor inthe concrete Operatiosn classes that took the model type from the generated code as a parameter

m-nash (approved):

PR #67, opened 2020-12-11
Update IDeletableResource Interface and Associated classes #4883

bquantump:


azure-proto-compute/AvailabilitySetOperations.cs
```diff
@@ -94,7 +103,7 @@ public Task<ArmOperation<AvailabilitySet>> AddTagAsync(string key, string value,
             return UpdateAsync(patchable);
         }
 
-        internal AvailabilitySetsOperations Operations => GetClient<ComputeManagementClient>((uri, cred) => new ComputeManagementClient(uri, Id.Subscription, cred, 
+        internal AvailabilitySetsOperations Operations => GetClient<ComputeManagementClient>((uri, cred) => new ComputeManagementClient(uri, Id.Subscription, cred,
```
allenjzhang, line 106: nit. Line returns to limit max 120.

markcowl (approved):

azure-proto-core/DeletableResourceOperations.cs
```diff
@@ -11,8 +11,12 @@ public interface IDeletableResource<TOperations, TResource>
         where TResource : Resource
```
m-nash, line 11: Can we make the file name and the interface name match?

m-nash (changes requested):

m-nash (approved):

PR #71, opened 2020-12-14
Add UserAssignedIdentity and its tests

nisha-bhatia:


nisha-bhatia:
Screenshot 2020-12-14 004855

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,129 @@
...
+            if (!ClientId.HasValue && !other.ClientId.HasValue)
+                return 0;
+
+            if (!ClientId.HasValue)
+                return -1;
+
+            if (!other.ClientId.HasValue)
+                return 1;
+
+            if (ClientId.Value.CompareTo(other.ClientId.Value) == 1 && PrincipalId.Value.CompareTo(other.PrincipalId.Value) == 1)
```
bquantump, line 38: Avoid these being called 2 times by calling it once and storing its value and using it in these two cases.

markcowl, line 38: I think this is a little difficult to understand logically.  I would prefer something like:

``` c#
var clientCompare = ClientId.Value.CompareTo(other.ClientId.Value);
if (clientCompare == 0)
{
    return PrincipalId.Value.CompareTo(other.PrincipalId.Value);
}

return clientCompare;
```

I think this is easier to inspect and determine that it is correct (if ClientIds are equal, only then does the principalId comparison make sense)

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,129 @@
...
+            PrincipalId = principalId;
+        }
+
+        public int CompareTo(UserAssignedIdentity other)
+        {
+            if (other == null)
+                return 1;
+
+            if (!ClientId.HasValue && !other.ClientId.HasValue)
+                return 0;
```
markcowl, line 30: What if ClientId is null and PrincipalId is not?  Is this a possible scenario? Or must PrincipalId have a value if ClientId does?

azure-proto-core-test/UserAssignedIdentityTests.cs
```diff
@@ -0,0 +1,377 @@
...
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+namespace azure_proto_core_test
+{
+    public class UserAssignedIdentityTests : UserAssignedIdentity
+    {
+        [TestCase(0, null, null, null, null)]
```
markcowl, line 12: Same comment as on SystemAssignedIdentities - make sure to exercise cases in which only one of the properties differs.

markcowl: Also, is it possible for client id or principal id to be null independently?

azure-proto-core-test/UserAssignedIdentityTests.cs
```diff
@@ -0,0 +1,377 @@
...
+                identity2 = new UserAssignedIdentity();
+
+            else
+                identity2 = new UserAssignedIdentity(new Guid(clientId2), new Guid(principalId2));
+
+            Assert.IsTrue(identity1.Equals(identity2));
+        }
+
+        [TestCase(null, null, "72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98")]
+        [TestCase("72f988bf-86f1-41af-91ab-2d7cd011db47", "de29bab1-49e1-4705-819b-4dfddceaaa98", "72f988bf-86f1-41af-91ab-2d7cd011db44", "de29bab1-49e1-4705-819b-4dfddceaaa94")]
```
markcowl, line 59: same comment as above - need to exercise cases where only one of the properties differs

azure-proto-core-test/UserAssignedIdentityTests.cs
```diff
@@ -0,0 +1,377 @@
+﻿using azure_proto_core.Resources;
+using NUnit.Framework;
+using System;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+namespace azure_proto_core_test
+{
+    public class UserAssignedIdentityTests : UserAssignedIdentity
```
markcowl, line 10: Per my comment on SystemAssignedIdentities, you should avoid having this class extend the class it is testing.  If you need a custom implementation of that class, consider having a test artifact used by this class.

azure-proto-core-test/UserAssignedIdentityTests.cs
```diff
@@ -0,0 +1,377 @@
...
+            }
+            string actual = "{\"clientId\":\"3beb288c-c3f9-4300-896f-02fbf175b6be\"," +
+                "\"principalId\":\"d0416856-d6cf-466d-8d64-ddc8d7782096\"}" +
+                "{\"clientId\":\"fbb39377-ff46-4a82-8c47-42d573180482\"," +
+                "\"principalId\":\"6d63ce43-c3ac-4b03-933d-4bc31eae50b2\"}";
+            Assert.AreEqual(actual, value);
+        }
+
+        [TestCase]
+        public void TestEqualsBothIdentitiesNull()
```
markcowl, line 330: Not sure how these differ from equality tests above.

markcowl (changes requested):

azure-proto-core/Resources/UserAssignedIdentity.cs
nisha-bhatia: the UserAssignedIdentity constructor doesn't allow for (Guid, null) because a Guid cannot be null. So we shouldn't ever have a case where ClientId has a value but PrincipalId does not

nisha-bhatia:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4467

azure-proto-core-test/UserAssignedIdentityTests.cs
nisha-bhatia: this is a static Equals function that can be called on two null identities. UserAssignedIdentity.Equals(null,null) is valid, but calling identity1.Equals(identity2) would not work

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
...
+        {
+            if (other == null)
+                return false;
+
+            return ClientId.Equals(other.ClientId) && PrincipalId.Equals(other.PrincipalId);
+        }
+
+        public static UserAssignedIdentity Deserialize(JsonElement element)
+        {
+            if (element.ValueKind == JsonValueKind.Undefined)
```
m-nash, line 49: I may have missed this in system identity, but I believe we should have 2 exceptions here
```suggestion
            if (element == null)
                throw new ArgumentNullException(nameof(element));
                
            if (element.ValueKind == JsonValueKind.Undefined)
                throw new ArgumentException("JsonElement cannot be undefined", nameof(element)); 
```

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
...
+            return new UserAssignedIdentity(clientId, principalId);
+        }
+
+        public static void Serialize(Utf8JsonWriter writer, UserAssignedIdentity userAssignedIdentity)
+        {
+            if (userAssignedIdentity == null)
+                throw new ArgumentNullException(nameof(userAssignedIdentity));
+
+            if (writer == null)
+                throw new ArgumentNullException("Utf8JsonWriter writer is null");
```
m-nash, line 88: use nameof like you did on line 85

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
...
+
+namespace azure_proto_core.Resources
+{
+    using System;
+    using System.Text.Json;
+    using Azure.Core;
+
+    public class UserAssignedIdentity
+    {
+        public Guid? ClientId { get; set; }
```
m-nash, line 12: should these be Guid?

I thought you always had these for UserAssignedIdentity.

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        public Guid? PrincipalId { get; set; }
```
m-nash, line 14: shouldn't we have the setter be private?

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace azure_proto_core.Resources
+{
+    using System;
```
m-nash, line 6: Move these outside the namespace, it was an old stylecop thing that was asking for these to be inside

azure-proto-core/Resources/UserAssignedIdentity.cs
```diff
@@ -0,0 +1,118 @@
...
+    using System.Text.Json;
+    using Azure.Core;
+
+    public class UserAssignedIdentity
+    {
+        public Guid? ClientId { get; set; }
+
+        public Guid? PrincipalId { get; set; }
+
+        public UserAssignedIdentity() { }
```
m-nash, line 16: is this public empty constructor needed / used?

m-nash (changes requested):

m-nash (approved):

markcowl (approved):

nisha-bhatia:
Updated ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4999

PR #73, opened 2020-12-15
Subscription model #4484

bquantump:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4484


Azure.ResourceManager.Core/Subscription.cs
```diff
@@ -0,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Azure.ResourceManager.Core
+{
+    public class Subscription : SubscriptionOperations
```
m-nash, line 6: Can you add a scenario test that spits out some other property of the subscription data other than Id like DisplayName just so we know end to end its working

m-nash (changes requested):

m-nash (approved):

PR #76, opened 2020-12-16
Location

JonathanCrd:
Implementation of Location methods and tests for that class.


Azure.ResourceManager.Core.Tests/LocationTests.cs
```diff
@@ -1,110 +1,173 @@
...
+        [TestCase(" ", " ", " ", " ")]
+        [TestCase(null, null, null, null)]
+        public void CanCreateLocation(string name, string expectedName, string expectedCanonincalName, string expectedDisplayName)
         {
-            Location location1 = new Location("West US");
-            Assert.IsFalse(location1.Equals("WestUS"));
+            Location location = name;
+            if (name == null)
+            {
+                Assert.Throws<System.NullReferenceException>(() => { string x = location.Name; });
```
m-nash, line 24: this isn't needed because its not testing any lines of code from Location.cs

Instead just Assert.IsNull(location);

Azure.ResourceManager.Core.Tests/LocationTests.cs
```diff
@@ -1,110 +1,173 @@
...
-            Location location1 = new Location("South US");
-            Assert.AreEqual(0, location1.CompareTo("South US"));
+            Location loc = location;
+            Assert.IsTrue(loc.DisplayName == location && loc.Name != location && loc.CanonicalName != location);
         }
 
-        [Test]
-        public void CanCompareToLocationGreaterThan()
+        [TestCase(true, "West Us", "West Us")]
+        [TestCase(true, "West Us", "WestUs")]
```
m-nash, line 78: Add a true case with special characters and one with numbers

Azure.ResourceManager.Core.Tests/LocationTests.cs
```diff
@@ -1,110 +1,173 @@
...
+            else
+            {
+                Assert.AreNotEqual(0, loc1.CompareTo(loc2));
+            }
         }
 
-        [Test]
-        public void CanCompareToStringGreaterThan()
+        [TestCase(true, "West Us", "West Us")]
+        [TestCase(true, "West Us", "WestUs")]
```
m-nash, line 101: same comment as above add true case for numbers and special characters

Azure.ResourceManager.Core.Tests/LocationTests.cs
```diff
@@ -1,110 +1,173 @@
...
+        [TestCase("South US", "West US", -1)]
+        [TestCase("South US", "west-us", -1)]
+        [TestCase("South US", null, 1)]
+        public void CompareToObject(string left, string right, int expected)
         {
-            Location location1 = new Location("South US");
-            Assert.AreEqual(-1, location1.CompareTo("West US"));
+            Location location1 = left;
+            Location location2 = right;
+            Assert.AreEqual(expected, location1.CompareTo(location2));
```
m-nash, line 148: Add the reverse check like you do with CompareToString

            if (right != null)
            {
                location1 = right;
                Assert.AreEqual(expected * -1, location1.CompareTo(left));
            }

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
     /// <summary>
-    ///     TODO: follow the full guidelines for these immutable types (IComparable, IEquatable, operator overloads, etc.)
+    /// Represents an Azure geography region where supported resource providers live.
     /// </summary>
-    public class Location : IEquatable<Location>, IEquatable<string>, IComparable<Location>, IComparable<string>
+    public class Location : IEquatable<Location>, IComparable<Location>
     {
-        public static readonly Location WestUS = new Location
-            { Name = "WestUS", CanonicalName = "west-us", DisplayName = "West US" };
+        /// <summary>
```
m-nash, line 16: Add a region around these static location objects

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
-        internal Location()
+        private enum NameType
         {
+            Name,
+            CanonicalName,
+            DisplayName,
         }
 
+        /// <summary>
+        /// Gets defaul Location object: West US.
```
m-nash, line 294: spelling default

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
+        /// Gets a location display name consisting of titlecase words or alphanumeric characters separated by whitespaces, e.g. "West US"
+        /// </summary>
+        public string DisplayName { get; private set; }
 
         /// <summary>
+        /// Creates a new location implicitly from a string.
         /// </summary>
-        /// <param name="other"></param>
-        public static implicit operator string(Location other)
+        /// <param name="other">String to be assigned.</param>
```
m-nash, line 316: mention it should be in the Name, CanonicalName, or DisplayName form

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
+
+            return new Location(other);
         }
 
-        public static implicit operator Location(string other)
+        /// <summary>
+        /// Creates a string implicitly from a Location object.
+        /// </summary>
+        /// <param name="other">Location object to be assigned.</param>
+        public static implicit operator string(Location other) => other.DisplayName;
```
m-nash, line 336: Can we point this at other.ToString() that way these are consistent in case we change the behavior

We also should need a null check here which if the unit tests pass probably means we are missing one.
if(other == null)
   return null;

Location x = null;
string y = x; //this currentlyresults in a null reference exception
Assert.IsNull(y);

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
-        {
-            return CanonicalName == other.CanonicalName;
+            return sb.ToString().ToUpperInvariant();
         }
 
-        public bool Equals(string other)
+        private static NameType DetectNameType(string location)
         {
-            return CanonicalName == GetCanonicalName(other);
+            string canonicalPattern = "^[a-z]+(-[a-z]+)+(-[1-9])?$";
```
m-nash, line 397: Can we make these two regex static readonly fields?

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,445 @@
...
 
-        private static string GetDefaultName(string name)
+        private static string GetDefaultNameFromDisplayName(string name)
         {
-            return name;
+            return Regex.Replace(name, @" ", string.Empty).ToLower();
         }
     }
-}
+}
```
m-nash, line 446: missing newline

m-nash (changes requested):

Azure.ResourceManager.Core/Resources/Location.cs
JonathanCrd: No, it says that the static and readonly keywords are not valid for those items.

Another possible way to do it is to declare them at the class level, would that be the way to go?

JonathanCrd:
Visual Studio - code coverage:
image

Cobertura Analysis:
image

Azure.ResourceManager.Core/Resources/Location.cs
m-nash: yeah thats what i mean, fields means class variable which is where the private static readonly modifiers apply

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,456 @@
...
+            }
+
+            return new string(chName);
         }
 
-        private static string GetDisplayName(string name)
+        private static string GetDefaultNameFromCanonicalName(string name)
         {
-            return name;
+            return Regex.Replace(name, @"-", string.Empty);
```
m-nash, line 449: same comment on the private static readonly

m-nash (changes requested):

Azure.ResourceManager.Core/Resources/Location.cs
JonathanCrd: Got it

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,458 @@
...
-        public string CanonicalName { get; internal set; }
+            var normalizedName = NormalizationUtility(other);
+            Location value;
+            if (PublicCloudLocations.TryGetValue(normalizedName, out value))
+            {
+                return value;
+            }
 
-        public string DisplayName { get; internal set; }
+            return new Location(other);
```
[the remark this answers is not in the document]
JonathanCrd: Yes, for the preview we will let an exception occur in that case.
This will be address in a future work item to get the available locations for a subscription.

Azure.ResourceManager.Core.Tests/LocationTests.cs
```diff
@@ -1,124 +1,225 @@
...
         {
-            Location location1 = "West US";
-            Assert.AreEqual("West US", location1.DisplayName);
+            Assert.IsNotNull(Location.Default);
+            Location location = Location.Default;
+            Assert.IsTrue(location.Equals(Location.Default));
         }
     }
-}
+}
```
m-nash, line 225: missing newline

Azure.ResourceManager.Core/Resources/Location.cs
```diff
@@ -2,98 +2,458 @@
...
+            }
+
+            return new string(chName);
         }
 
-        private static string GetDisplayName(string name)
+        private static string GetDefaultNameFromCanonicalName(string name)
         {
-            return name;
+            return Regex.Replace(name, RegexDash, string.Empty);
```
m-nash, line 451: Can we see whats faster here
Regex.Replace(name, RegexDash, string.Empty)
vs
name.Replace(" ", string.Empty)

m-nash (changes requested):

Azure.ResourceManager.Core/Resources/Location.cs
JonathanCrd: Done, name.Replace definetly runs faster than Regex.Replace.
image

I'll replace all the Regex.Replace with the string Replace method.

m-nash (approved):

PR #85, opened 2020-12-18
Update all properties in Sku and Plan class to have private setters

YalinLi0312:


Azure.ResourceManager.Core/Resources/Plan.cs
```diff
@@ -10,15 +10,24 @@ namespace Azure.ResourceManager.Core
...
-        public string Product { get; set; }
+        public string Product { get; private set; }
 
-        public string PromotionCode { get; set; }
+        public string PromotionCode { get; private set; }
 
-        public string Version { get; set; }
+        public string Version { get; private set; }
+
+        public Plan(string name, string publisher, string product, string promotionCode, string version)
```
m-nash, line 23: this should be internal

Azure.ResourceManager.Core/Resources/Sku.cs
```diff
@@ -10,15 +10,24 @@ namespace Azure.ResourceManager.Core
...
-        public string Family { get; set; }
+        public string Family { get; private set; }
 
-        public string Size { get; set; }
+        public string Size { get; private set; }
 
-        public long? Capacity { get; set; }
+        public long? Capacity { get; private set; }
+
+        public Sku(string name, string tier, string family, string size, long? capacity = null)
```
m-nash, line 23: this should be internal

m-nash (changes requested):

m-nash (approved):

PR #90, opened 2020-12-30
Remove generic type parameters from ResourceListOperations.cs

nisha-bhatia:
ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4963

### DIFF
--- a/Azure.ResourceManager.Core/ResourceListOperations.cs
+++ b/Azure.ResourceManager.Core/ResourceListOperations.cs
@@ -11,7 +11,7 @@ using Azure.ResourceManager.Resources.Models;
 namespace Azure.ResourceManager.Core
 {
     /// <summary>
-    ///     Generic list operations class - can be extended if a specific RP has more list operations
+    ///  Generic list operations class. This can be extended if a specific RP has more list operations.
     /// </summary>
     public class ResourceListOperations
     {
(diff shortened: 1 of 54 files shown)


Azure.ResourceManager.Core/ResourceListOperations.cs
```diff
@@ -15,20 +15,20 @@ namespace Azure.ResourceManager.Core
     /// </summary>
     public class ResourceListOperations
     {
-        public static Pageable<TOperations> ListAtContext<TOperations, TResource>(
+        public static Pageable<ArmResourceOperations> ListAtContext<ArmResourceOperations, ArmResource>(
```
m-nash, line 18: I think we can remove the generic type completely
```suggestion
        public static Pageable<ArmResourceOperations> ListAtContext(
```

m-nash (changes requested):

Azure.ResourceManager.Core/ResourceListOperations.cs
markcowl: If these methods are meant simply to call the generic ARM APIs, then the proper return type would be 'ArmResource', because it will have the ARM envelope as well as the operations, and then no generic parameters are needed at all, as Michael suggests.

I am assuming that this would then be used by generated list methods in the Collection classes to transform the ```Pageable<ArmResource>``` into ```Pageable<MyResourceOperations>``` when presenting a method to the user like ```Pageable<VirtualMachineOperations> List(ArmFilterCollection, int? CancellationToken)```

markcowl (changes requested):

m-nash (approved):

markcowl (approved):

PR #91, opened 2021-01-05
Remove ResourceType property from the OperationsBase class

YalinLi0312:
ADO workitem: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4964

### DIFF
--- a/Azure.ResourceManager.Core/SubscriptionContainer.cs
+++ b/Azure.ResourceManager.Core/SubscriptionContainer.cs
@@ -12,8 +12,6 @@ namespace Azure.ResourceManager.Core
     /// </summary>
     public class SubscriptionContainer : OperationsBase
     {
-        public static readonly ResourceType ResourceType = "Microsoft.Resources/subscriptions";
-
         internal SubscriptionContainer(AzureResourceManagerClientOptions options)
             : base(options, null, null)
         {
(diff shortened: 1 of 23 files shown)


Azure.ResourceManager.Core/SubscriptionContainer.cs
```diff
@@ -12,15 +12,13 @@ namespace Azure.ResourceManager.Core
     /// </summary>
     public class SubscriptionContainer : OperationsBase
     {
-        public static readonly ResourceType AzureResourceType = "Microsoft.Resources/subscriptions";
+        public static readonly ResourceType ResourceType = "Microsoft.Resources/subscriptions";
```
m-nash, line 15: I think we should remove these from the container classes and only have it on the operations class

m-nash (changes requested):

Azure.ResourceManager.Core/SubscriptionContainer.cs
YalinLi0312: Agree, it a duplicate definition. Will update soon.

m-nash (approved):

PR #96, opened 2021-01-07
Adding Subscription XmlDoc

allenjzhang:


Azure.ResourceManager.Core/SubscriptionContainer.cs
```diff
@@ -7,40 +7,74 @@
...
+        /// </summary>
+        internal SubscriptionsOperations Operations => GetClient((uri, cred) =>
+            new ResourcesManagementClient(
+                uri,
+                Guid.NewGuid().ToString(),
+                cred,
+                ClientOptions.Convert<ResourcesManagementClientOptions>())).Subscriptions;
 
+        /// <summary>
+        /// List all subscriptions in the current container.
```
YalinLi0312, line 34: I think if we decide to use `Gets` in above, then here should use `Lists`.

PR #98, opened 2021-01-08
Proposal for naming changes for operations/container extensions

weshaggard:
In some cases it might seem like we should use properties but
given our desire to break dependencies we cannot use properties
because we have to use extension methods which don't support properties.

On top of that we should probably avoid properties as well because
these are more then simple data field accesses. They always construct
a new wrapper object that gets returned and based on some of the
standard guidelines we should use methods

See https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229054(v=vs.100)?redirectedfrom=MSDN)
"The operation returns a copy of an internal state (this does not include copies of value type objects returned on the stack)."

For those reasons we should use method for both the instance and extension methods
and so it now comes down to how we name these methods.

Currently we have methods that look like:

Subscription() -> return new SubscriptionOperations
Subscriptions() -> returns new SubscriptionContainer
ResourceGroup() -> return new ResourceGroupOperations
ResourceGroups() -> return new ResourceGroupContainer
VirtualMachine() -> returns new VirtualMachineOperations or VirtualMachine
VirtualMachines() -> returns new VirtualMachineContainer
AvailabilitySet() -> return new AvailabilitySetOperations or AvailabilitySet
AvailabilitySets() -> return new AvailabilitySetContainer

Proposal for method name changes:

Based on naming guidelines (https://docs.microsoft.com/en-us/previous-versions/dotnet/netframework-4.0/ms229012(v=vs.100)#names-of-methods)
for methods we shoudl use a verb describing wht the method action is doing. So in our
cases we are "Getting" stuff so the suggestion is to prefix the methods with "Get".

On top of that it isn't very clear of what object we are getting in different contexts so
we the second part of the proposal is to name the method after the type that will be
returned.

GetSubscriptionOperations() -> return new SubscriptionOperations
GetSubscriptionContainer() -> returns new SubscriptionContainer
GetResourceGroupOperations() -> return new ResourceGroupOperations
GetResourceGroupContainer() -> return new ResourceGroupContainer
GetVirtualMachineOperations() -> returns new VirtualMachineOperations
GetVirtualMachine() -> returns new VirtualMachine
GetVirtualMachineContainer() -> returns new VirtualMachineContainer
GetAvailabilitySetOperations() -> return new AvailabilitySetOperations
GetAvailabilitySet() -> return new AvailabilitySet
GetAvailabilitySetContainer() -> return new AvailabilitySetContainer

This a very basic design that makes it clear what you are getting and also
gives people an indication that they might want to store the value if they
plan to make many requests on the returned object. The names might be a little
long but the pattern of the simple mapping between the name and the what you are
expecting the method to do very clear.

### DIFF
--- a/Azure.ResourceManager.Core/AzureResourceManagerClient.cs
+++ b/Azure.ResourceManager.Core/AzureResourceManagerClient.cs
@@ -128,29 +128,22 @@ namespace Azure.ResourceManager.Core
         /// <summary>
         /// Gets the Azure subscription operations.
         /// </summary>
-        /// <param name="subscription">  The data model of the subscription. </param>
+        /// <param name="subscriptionData">  The data model of the subscription. </param>
         /// <returns> Subscription operations. </returns>
-        public SubscriptionOperations Subscription(SubscriptionData subscription) => new SubscriptionOperations(ClientOptions, subscription);
+        public Subscription GetSubscription(SubscriptionData subscriptionData) => new Subscription(ClientOptions, subscriptionData);
 
         /// <summary>
         /// Gets the Azure subscription operations.
         /// </summary>
-        /// <param name="subscription"> The resource identifier of the subscription. </param>
+        /// <param name="subscriptionId"> The resource identifier of the subscription. </param>
         /// <returns> Subscription operations. </returns>
-        public SubscriptionOperations Subscription(ResourceIdentifier subscription) => new SubscriptionOperations(ClientOptions, subscription);
-
-        /// <summary>
-        /// Gets the Azure subscription operations.
-        /// </summary>
-        /// <param name="subscription"> The id of the subscription. </param>
-        /// <returns> Subscription operations. </returns>
-        public SubscriptionOperations Subscription(string subscription) => new SubscriptionOperations(ClientOptions, subscription);
+        public SubscriptionOperations GetSubscriptionOperations(ResourceIdentifier subscriptionId) => new SubscriptionOperations(ClientOptions, subscriptionId);
 
         /// <summary>
         /// Gets the Azure subscriptions.
         /// </summary>
         /// <returns> Subscription container. </returns>
-        public SubscriptionContainer Subscriptions()
+        public SubscriptionContainer GetSubscriptionContainer()
         {
             return new SubscriptionContainer(ClientOptions);
         }
@@ -275,32 +268,32 @@ namespace Azure.ResourceManager.Core
         /// <summary>
         /// Gets resource group operations.
         /// </summary>
-        /// <param name="subscription"> The id of the Azure subscription. </param>
-        /// <param name="resourceGroup"> The resource group name. </param>
+        /// <param name="subscriptionGuid"> The id of the Azure subscription. </param>
+        /// <param name="resourceGroupName"> The resource group name. </param>
         /// <returns> Resource group operations. </returns>
-        public ResourceGroupOperations ResourceGroup(string subscription, string resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(string subscriptionGuid, string resourceGroupName)
         {
-            return new ResourceGroupOperations(ClientOptions, $"/subscriptions/{subscription}/resourceGroups/{resourceGroup}");
+            return new ResourceGroupOperations(ClientOptions, $"/subscriptions/{subscriptionGuid}/resourceGroups/{resourceGroupName}");
         }
 
         /// <summary>
         /// Gets resource group operations.
         /// </summary>
-        /// <param name="resourceGroup"> The resource identifier of the resource group. </param>
+        /// <param name="resourceGroupName"> The resource identifier of the resource group. </param>
         /// <returns> Resource group operations. </returns>
-        public ResourceGroupOperations ResourceGroup(ResourceIdentifier resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(ResourceIdentifier resourceGroupName)
         {
-            return new ResourceGroupOperations(ClientOptions, resourceGroup);
+            return new ResourceGroupOperations(ClientOptions, resourceGroupName);
         }
 
         /// <summary>
         /// Gets resource group operations.
         /// </summary>
-        /// <param name="resourceGroup"> The data model of the resource group. </param>
+        /// <param name="resourceGroupName"> The data model of the resource group. </param>
         /// <returns> Resource group operations. </returns>
-        public ResourceGroupOperations ResourceGroup(ResourceGroupData resourceGroup)
+        public ResourceGroup GetResourceGroup(ResourceGroupData resourceGroupData)
         {
-            return new ResourceGroupOperations(ClientOptions, resourceGroup.Id);
+            return new ResourceGroup(ClientOptions, resourceGroupData);
         }
 
         /// <summary>
@@ -385,7 +378,7 @@ namespace Azure.ResourceManager.Core
             string sub = DefaultSubscription?.Id?.Subscription;
             if (null == sub)
             {
-                sub = await Subscriptions().GetDefaultSubscriptionAsync(cancellationToken);
+                sub = await GetSubscriptionContainer().GetDefaultSubscriptionAsync(cancellationToken);
             }
 
             return sub;
--- a/Azure.ResourceManager.Core/SubscriptionOperations.cs
+++ b/Azure.ResourceManager.Core/SubscriptionOperations.cs
@@ -22,16 +22,6 @@ namespace Azure.ResourceManager.Core
         /// Initializes a new instance of the <see cref="SubscriptionOperations"/> class.
         /// </summary>
         /// <param name="options"> The client parameters to use in these operations. </param>
-        /// <param name="subscriptionId"> The Id of the subscription. </param>
-        internal SubscriptionOperations(AzureResourceManagerClientOptions options, string subscriptionId)
-            : base(options, $"/subscriptions/{subscriptionId}")
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="SubscriptionOperations"/> class.
-        /// </summary>
-        /// <param name="options"> The client parameters to use in these operations. </param>
         /// <param name="id"> The identifier of the subscription. </param>
         internal SubscriptionOperations(AzureResourceManagerClientOptions options, ResourceIdentifier id)
             : base(options, id)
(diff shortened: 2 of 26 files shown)


m-nash:
I like where you are going with this, after looking at this I see a few areas where we didn't update our return types.  As an example in AzureResourceManagerClient we have
`        public SubscriptionOperations Subscription(SubscriptionData subscription) => new SubscriptionOperations(ClientOptions, subscription);
`
technically this should have been updated to
`        public Subscription Subscription(SubscriptionData subscription) => new Subscription(ClientOptions, subscription);
`
Since in this case we have the full data object.  There might exist an overload that has
`        public SubscriptionOperations Subscription(ResourceIdentifier subscriptionId) => new SubscriptionOperations(ClientOptions, subscriptionId);
`
How would the naming adapt to these changes?

Azure.ResourceManager.Core/SubscriptionContainer.cs
```diff
@@ -7,7 +7,7 @@
 namespace Azure.ResourceManager.Core
 {
     /// <summary>
-    /// Subscription Container Operationss
+    /// GetSubscriptionOperations Container Operationss
```
m-nash, line 10: I think this comment updated wasn't intended?

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -71,17 +71,17 @@ public AzureResourceManagerClient(Uri baseUri, TokenCredential credential, strin
 
         internal virtual AzureResourceManagerClientOptions ClientOptions { get; }
 
-        public SubscriptionOperations Subscription(SubscriptionData subscription) => new SubscriptionOperations(ClientOptions, subscription);
+        public SubscriptionOperations GetSubscriptionOperations(SubscriptionData subscription) => new SubscriptionOperations(ClientOptions, subscription);
```
m-nash, line 74: should return Subscription instead of SubscriptionOperations
```suggestion
        public Subscription GetSubscription(SubscriptionData subscription) => new Subscription(ClientOptions, subscription);
```

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -71,17 +71,17 @@ public AzureResourceManagerClient(Uri baseUri, TokenCredential credential, strin
...
 
         /// <summary>
         /// </summary>
         /// <param name="subscription"></param>
         /// <returns></returns>
-        public SubscriptionOperations Subscription(ResourceIdentifier subscription) => new SubscriptionOperations(ClientOptions, subscription);
+        public SubscriptionOperations GetSubscriptionOperations(ResourceIdentifier subscription) => new SubscriptionOperations(ClientOptions, subscription);
```
m-nash, line 80: param should indicate that its an Id not an actual subscription
```suggestion
        public SubscriptionOperations GetSubscriptionOperations(ResourceIdentifier subscriptionId) => new SubscriptionOperations(ClientOptions, subscriptionId);
```

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -71,17 +71,17 @@ public AzureResourceManagerClient(Uri baseUri, TokenCredential credential, strin
...
 
-        public SubscriptionOperations Subscription(string subscription) => new SubscriptionOperations(ClientOptions, subscription);
+        public SubscriptionOperations GetSubscriptionOperations(string subscription) => new SubscriptionOperations(ClientOptions, subscription);
```
m-nash, line 82: We should remove this overload since there is an implicit conversion from string -> ResourceIdentifier

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -164,17 +164,17 @@ public IEnumerable<Location> ListAvailableLocations(string subscription, Resourc
                 .Locations.Cast<Location>();
         }
 
-        public ResourceGroupOperations ResourceGroup(string subscription, string resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(string subscription, string resourceGroup)
```
m-nash, line 167: params should be more explicit
```suggestion
        public ResourceGroupOperations GetResourceGroupOperations(string subscriptionGuid, string resourceGroupName)
```

weshaggard:
>I assume we would follow the same pattern you layed out for GetVirtualMachine for GetSubscription instead of GetSubscriptionOperations?

Yes exactly.

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -164,17 +164,17 @@ public IEnumerable<Location> ListAvailableLocations(string subscription, Resourc
...
         {
             return new ResourceGroupOperations(ClientOptions, $"/subscriptions/{subscription}/resourceGroups/{resourceGroup}");
         }
 
-        public ResourceGroupOperations ResourceGroup(ResourceIdentifier resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(ResourceIdentifier resourceGroup)
```
m-nash, line 172: param name changes
```suggestion
        public ResourceGroupOperations GetResourceGroupOperations(ResourceIdentifier resourceGroupId)
```

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -164,17 +164,17 @@ public IEnumerable<Location> ListAvailableLocations(string subscription, Resourc
...
         {
             return new ResourceGroupOperations(ClientOptions, resourceGroup);
         }
 
-        public ResourceGroupOperations ResourceGroup(ResourceGroupData resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(ResourceGroupData resourceGroup)
```
m-nash, line 177: return ResourceGroup
```suggestion
        public ResourceGroup GetResourceGroup(ResourceGroupData resourceGroup)
```

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -229,7 +229,7 @@ internal async Task<string> GetDefaultSubscription(CancellationToken cancellatio
             string sub = DefaultSubscription?.Id?.Subscription;
             if (null == sub)
             {
-                sub = await Subscriptions().GetDefaultSubscription(cancellationToken);
+                sub = await GetSubscriptionContainer().GetDefaultSubscription(cancellationToken);
```
m-nash, line 232: I think you are missing latest code here, sync with master to verify

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -9,7 +9,7 @@
 namespace Azure.ResourceManager.Core
 {
     /// <summary>
-    ///     Subscription operations
+    ///     GetSubscriptionOperations operations
```
m-nash, line 12: I think this comment update wasn't intended

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -36,22 +36,22 @@ internal SubscriptionOperations(AzureResourceManagerClientOptions options, Resou
         internal ResourceGroupsOperations RgOperations => GetClient<ResourcesManagementClient>((uri, cred) =>
             new ResourcesManagementClient(uri, Id.Subscription, cred, ClientOptions.Convert<ResourcesManagementClientOptions>())).ResourceGroups;
 
-        public ResourceGroupOperations ResourceGroup(ResourceGroupData resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(ResourceGroupData resourceGroup)
```
m-nash, line 39: return ResourceGroup
```suggestion
        public ResourceGroup GetResourceGroup(ResourceGroupData resourceGroup)
```

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -36,22 +36,22 @@ internal SubscriptionOperations(AzureResourceManagerClientOptions options, Resou
...
         {
             return new ResourceGroupOperations(ClientOptions, resourceGroup);
         }
 
-        public ResourceGroupOperations ResourceGroup(ResourceIdentifier resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(ResourceIdentifier resourceGroup)
```
m-nash, line 44: param rename
```suggestion
        public ResourceGroupOperations GetResourceGroupOperations(ResourceIdentifier resourceGroupId)
```

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -36,22 +36,22 @@ internal SubscriptionOperations(AzureResourceManagerClientOptions options, Resou
...
         {
             return new ResourceGroupOperations(ClientOptions, resourceGroup);
         }
 
-        public ResourceGroupOperations ResourceGroup(string resourceGroup)
+        public ResourceGroupOperations GetResourceGroupOperations(string resourceGroup)
```
m-nash, line 49: param rename
```suggestion
        public ResourceGroupOperations GetResourceGroupOperations(string resourceGroupName)
```

azure-proto-compute/Convenience/VirtualMachineModelBuilder.cs
```diff
@@ -10,7 +10,7 @@ public class VirtualMachineModelBuilder : VirtualMachineModelBuilderBase
         internal VirtualMachineModelBuilder(VirtualMachineContainer containerOperations, VirtualMachineData vm): base(containerOperations, vm)
         {
             // _model.Name = vmName;
-            //_model = new VirtualMachine(location);
+            //_model = new GetVirtualMachineOperations(location);
```
m-nash, line 13: undo comment change

m-nash (changes requested):

Azure.ResourceManager.Core/SubscriptionContainer.cs
weshaggard: No that was the automate rename refactoring that updated this comment incorrectly.

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -8,45 +8,24 @@ namespace azure_proto_compute
     public static class ResourceGroupExtensions
     {
         #region VirtualMachines
-        /// <summary>
-        /// Gets an object representing the operations that can be performed over a specific VirtualMachine.
-        /// </summary>
-        /// <param> The <see cref="[ResourceGroupOperations]" /> instance the method will execute against. </param>
-        /// <param name="vmName"> The name of the VirtualMachine. </param>
-        /// <returns> Returns an object representing the operations that can be performed over a specific <see cref="[VirtualMachine]" />.</returns>
-        public static VirtualMachineOperations VirtualMachine(this ResourceGroupOperations resourceGroup, string vmName)
```
m-nash, line 17: undo this removal

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -56,51 +35,31 @@ public static VirtualMachineContainer VirtualMachines(this ResourceGroup resourc
...
         }
         #endregion
 
         #region AvailabilitySets
-        /// <summary>
-        /// Gets an object representing the operations that can be performed over a specific AvailabilitySet.
-        /// <param> The <see cref="[ResourceGroupOperations]" /> instance the method will execute against. </param>
-        /// <param name="availabilitySetName"> The name of the AvailibilitySet. </param>
-        /// <returns> Returns an object representing the operations that can be performed over a specific <see cref="[AvailabilitySet]" />. </returns>
-        public static AvailabilitySetOperations AvailabilitySet(this ResourceGroupOperations resourceGroup, string availabilitySetName)
```
m-nash, line 71: undo this removal

m-nash (changes requested):

m-nash (approved):

PR #109, opened 2021-01-11
Pr for #4465

bquantump:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4465

### DIFF
--- a/Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
+++ b/Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
@@ -1,6 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
-
+using System;
 namespace Azure.ResourceManager.Core
 {
     /// <summary>
@@ -19,10 +19,12 @@ namespace Azure.ResourceManager.Core
 
         protected TrackedResource(string id, Location location, T data)
         {
-            if(object.ReferenceEquals(id, null)){
+            if (object.ReferenceEquals(id, null))
+            {
                 Id = null;
             }
-            else{
+            else
+            {
                 Id = id;
             }
             Location = location;
--- a/Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
+++ b/Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
@@ -17,6 +17,18 @@ namespace Azure.ResourceManager.Core
             Model = data;
         }
 
+        protected TrackedResource(string id, Location location, T data)
+        {
+            if(object.ReferenceEquals(id, null)){
+                Id = null;
+            }
+            else{
+                Id = id;
+            }
+            Location = location;
+            Model = data;
+        }
+
         public virtual T Model { get; set; }
 
         public static implicit operator T(TrackedResource<T> other)
--- a/Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
+++ b/Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
@@ -52,8 +52,7 @@ namespace Azure.ResourceManager.Core
 
         public static implicit operator ResourceIdentifier(string other)
         {
-            var val = other == null ? throw new ArgumentNullException(nameof(other)) : new ResourceIdentifier(other);
-            return val;
+            return new ResourceIdentifier(other); // will null check in PR #119
         }
 
         /// <summary>
@@ -141,10 +140,6 @@ namespace Azure.ResourceManager.Core
         /// <param name="id">A properly formed resource identity</param>
         protected virtual void Parse(string id)
         {
-            if (id == null)
-            {
-                throw new ArgumentNullException(nameof(id));
-            }
             // Throw for null, empty, and string without the correct form
             if (string.IsNullOrWhiteSpace(id) || !id.Contains('/'))
                 throw new ArgumentOutOfRangeException($"'{id}' is not a valid resource");
--- a/Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
+++ b/Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
@@ -21,10 +21,7 @@ namespace Azure.ResourceManager.Core
         public ResourceIdentifier(string id)
         {
             Id = id;
-
-            //TODO: Due to the implicit this is called for blank constructions such as new PhResourceGroup
-            if (id != null)
-                Parse(id);
+            Parse(id);
         }
 
         public string Id { get; protected set; }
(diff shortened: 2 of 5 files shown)


Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
```diff
@@ -17,6 +17,18 @@ protected TrackedResource(ResourceIdentifier id, Location location, T data)
...
+            if(object.ReferenceEquals(id, null)){
+                Id = null;
+            }
+            else{
+                Id = id;
+            }
+            Location = location;
+            Model = data;
+        }
+
```
bquantump, line 31: Why not simplify this to Id = object.ReferenceEquals(id, null)  ? null : id;?

This fails interestingly enough. 

The CIL ouput differs in that it puts one thing on the allocation stack and the set it at the end, making the runtime have to think whatever will be set will be a string, causing the unwanted implicit this to be call.


image

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -60,7 +57,8 @@ public override int GetHashCode()
 
         public static implicit operator ResourceIdentifier(string other)
         {
-            return new ResourceIdentifier(other);
+            var val = other == null ? throw new ArgumentNullException(nameof(other)) : new ResourceIdentifier(other);
```
m-nash, line 60: for style reasons can we change this to an if throw otherwise assign?
```suggestion
            if(other == null)
                throw new ArgumentNullException(nameof(other));
            var val = new ResourceIdentifier(other);
```

Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
```diff
@@ -17,6 +17,20 @@ protected TrackedResource(ResourceIdentifier id, Location location, T data)
             Model = data;
         }
 
+        protected TrackedResource(string id, Location location, T data)
```
m-nash, line 20: need some context as to why this is needed.  Is this to avoid the automatic implicit call, allowing us to have null as the id for "request" objects?

m-nash (changes requested):

Azure.ResourceManager.Core/Adapters/ModelAdapters.cs
bquantump: Exactly, if null get's passed into the other constructor, the implicit this get's called. This allows us to properly avoid that and set null for request types

client/Program.cs
```diff
@@ -10,7 +10,7 @@ static void Main(string[] args)
             Scenario scenario = null;
             try
             {
-                scenario = ScenarioFactory.GetScenario(Scenarios.All);
```
allenjzhang, line 13: nit: Revert this change

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -21,10 +21,7 @@ public class ResourceIdentifier : IEquatable<ResourceIdentifier>, IEquatable<str
         public ResourceIdentifier(string id)
         {
             Id = id;
```
allenjzhang, line 23: let's add null check/verification at any public method entry:               if (string.IsNullOrWhiteSpace(id)) throw new ArgumentNullException(nameof(id));

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -143,6 +141,10 @@ public override string ToString()
         /// <param name="id">A properly formed resource identity</param>
         protected virtual void Parse(string id)
```
allenjzhang, line 142: >protected virtual [](start = 8, length = 17)

private? Why protected virtual? is ResourceIdentifier to be sub-classed?

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -143,6 +141,10 @@ public override string ToString()
...
         {
+            if (id == null)
```
allenjzhang, line 144: if this is private, these checks should be at public method entry point.

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -143,6 +141,10 @@ public override string ToString()
...
+            {
+                throw new ArgumentNullException(nameof(id));
+            }
             // Throw for null, empty, and string without the correct form
             if (string.IsNullOrWhiteSpace(id) || !id.Contains('/'))
```
allenjzhang, line 149: I think we can use a simple regex to do validation:  /.*

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
```diff
@@ -143,6 +141,10 @@ public override string ToString()
...
                 throw new ArgumentOutOfRangeException($"'{id}' is not a valid resource");
```
allenjzhang, line 150: Should be ArgumentException.

allenjzhang:
>             if (!(KnownKeys.Subscription.Equals(parts[0], StringComparison.InvariantCultureIgnoreCase) ||

Added a discussion point on implementation in Sync meeting for 1/13.

---
Refers to: Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs:161 in 729d3c1. [](commit_id = 729d3c13bb7caad4ed9ce473f9aa867ccb651ba2, deletion_comment = False)

allenjzhang:
>             public static string UntrackedSubResource => Guid.NewGuid().ToString();

I am not sure that untracked resources are all GUID.

---
Refers to: Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs:250 in 729d3c1. [](commit_id = 729d3c13bb7caad4ed9ce473f9aa867ccb651ba2, deletion_comment = False)

allenjzhang:
>         public static class KnownKeys

private?

---
Refers to: Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs:238 in 729d3c1. [](commit_id = 729d3c13bb7caad4ed9ce473f9aa867ccb651ba2, deletion_comment = False)

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
bquantump: Null check will be added when we decided on #118

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
bquantump: Bring up at meeting in general to go over parsing ID

m-nash (changes requested):
Can you add tests which now verify the original issue no longer exists?

m-nash (approved):

Azure.ResourceManager.Core/Resources/ResourceIdentifier.cs
bquantump: I will add null checks with the outcome of #118

PR #113, opened 2021-01-12
Update XML documentation in AvailabilitySet to follow the latest standard

YalinLi0312:


azure-proto-compute/AvailabilitySetOperations.cs
```diff
@@ -165,6 +163,9 @@ public async Task<ArmOperation<AvailabilitySet>> UpdateAsync(AvailabilitySetUpda
         /// </summary>
```
m-nash, line 163: use standard summary text
https://github.com/m-nash/azure-proto-sdk/blob/master/XmlDocStandard.md#tags

azure-proto-compute/AvailabilitySetOperations.cs
```diff
@@ -178,6 +179,9 @@ public ArmOperation<AvailabilitySet> StartAddTag(string key, string value)
...
```
m-nash, line 179: use standard summary text
https://github.com/m-nash/azure-proto-sdk/blob/master/XmlDocStandard.md#tags

m-nash (changes requested):

m-nash (approved):

PR #122, opened 2021-01-13
Add XML Documentation to ResourceGroupExtensions.cs

nisha-bhatia:
ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=4897

### DIFF
--- a/azure-proto-compute/Extensions/ResourceGroupExtensions.cs
+++ b/azure-proto-compute/Extensions/ResourceGroupExtensions.cs
@@ -12,7 +12,7 @@ namespace azure_proto_compute
         /// Gets an object representing the operations that can be performed over a specific VirtualMachine.
         /// </summary>
         /// <param> The <see cref="[ResourceGroupOperations]" /> instance the method will execute against. </param>
-        /// <param name="vmName"> The name of the <see cref="[VirtualMachine]" />. </param>
+        /// <param name="vmName"> The name of the VirtualMachine. </param>
         /// <returns> Returns an object representing the operations that can be performed over a specific <see cref="[VirtualMachine]" />.</returns>
         public static VirtualMachineOperations VirtualMachine(this ResourceGroupOperations resourceGroup, string vmName)
         {
@@ -31,10 +31,10 @@ namespace azure_proto_compute
         }
 
         /// <summary>
-        /// Gets a VirtualMachine.
+        /// Gets an object representing a VirtualMachine along with the instance operations that can be performed on it.
         /// </summary>
         /// <param> The <see cref="[ResourceGroup]" /> instance the method will execute against. </param>
-        /// <param name="vmData"> The <see cref="[VirtualMachineData]" /> of the resource that is the target of operations. </param>
+        /// <param name="vmData"> The VirtualMachineData of the resource that is the target of operations. </param>
         /// <returns> Returns a <see cref="[VirtualMachine]" /> object. </returns>
         public static VirtualMachine VirtualMachine(this ResourceGroup resourceGroup, VirtualMachineData vmData)
         {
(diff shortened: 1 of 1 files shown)


azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -3,58 +3,114 @@
...
     /// <summary>
-    /// Naming of these methods
+    /// A class to add extension methods to ResourceGroup.
     /// </summary>
     public static class ResourceGroupExtensions
     {
         #region VirtualMachines
-        public static VirtualMachineOperations VirtualMachine(this ResourceGroupOperations resourceGroup, string name)
+        /// <summary>
+        /// Gets the VirtualMachineOperations for a VirtualMachine.
```
m-nash, line 12: I think it might work out well to use the same phrasing we do on the class description for VirtualMachineOperations just put a Gets in front of it

https://github.com/m-nash/azure-proto-sdk/blob/master/XmlDocStandard.md#standard-class-documentation

so here it would read
```suggestion
        /// Gets an object representing the operations that can be performed over a specific VirtualMachine.
```

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -3,58 +3,114 @@
...
+        /// </summary>
+        /// <param> The <see  cref="[ResourceGroupOperations]" /> instance the method will execute against. </param>
+        /// <param name="vmName"> The name of the VirtualMachine. </param>
+        /// <returns> Returns a response with the <see cref="VirtualMachineOperations"/> operation for this resource. </returns>
```
m-nash, line 16: The "Returns a response" wording is slightly out of context here.  It comes from here https://github.com/m-nash/azure-proto-sdk/blob/master/XmlDocStandard.md#method which is referring to an http response.

Here we are simply returning an instance and we might use the class text again with a small modification
https://github.com/m-nash/azure-proto-sdk/blob/master/XmlDocStandard.md#standard-class-documentation

```suggestion
        /// <returns> Returns an object representing the operations that can be performed over a specific VirtualMachine.</returns>
```

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -3,58 +3,114 @@
...
         {
             return new VirtualMachineOperations(resourceGroup.ClientOptions, vmId);
         }
 
-        public static VirtualMachine VirtualMachine(this ResourceGroup resourceGroup, VirtualMachineData vm)
+        /// <summary>
+        /// Gets the VirtualMachineOperations for a VirtualMachine.
+        /// </summary>
+        /// <param> The <see  cref="[ResourceGroup]" /> instance the method will execute against. </param>
+        /// <param name="vmData"> The VirtualMachineData of the resource that is the target of operations. </param>
```
m-nash, line 37: use a ref link here
```suggestion
        /// <param name="vmData"> The <see cref="VirtualMachineData" /> of the resource that is the target of operations. </param>
```

m-nash (changes requested):

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -9,53 +9,53 @@ public static class ResourceGroupExtensions
     {
         #region VirtualMachines
         /// <summary>
-        /// Gets the VirtualMachineOperations for a VirtualMachine.
+        /// Gets an object representing the operations that can be performed over a specific <see cref="[VirtualMachine]" />.
```
m-nash, line 12: the cref should be VirtualMachineOperations since thats what is being returned, or it should simply just have text VirtualMachine.  I would lean towards just text since the type is already in the response statement

m-nash (changes requested):

m-nash (approved):

PR #123, opened 2021-01-14
Workshop

m-nash:


allenjzhang:
>         /// <param> The <see cref="[SubscriptionOperations]" /> instance the method will execute against. </param>

There are bunch of incorrectly labelled crefs with extra []. You can do a global search/replace with these regex: cref=\"\[(.*)\]"   --> cref="$1"

---
Refers to: azure-proto-compute/Extensions/SubscriptionExtensions.cs:43 in 2419945. [](commit_id = 2419945df2b650c11899736160e069770baa4f7e, deletion_comment = False)

Azure.ResourceManager.Core/Adapters/PhWrappingAsyncPageable.cs
```diff
@@ -3,135 +3,58 @@
...
-using System.Collections.Immutable;
 using System.Linq;
 using System.Threading.Tasks;
-using Azure;
 
 namespace Azure.ResourceManager.Core.Adapters
 {
     /// <summary>
-    ///     Returns an AsyncPageable that executes a given task before retrieving the first page of results
+    /// As above, returns an AsyncPageable that transforms each page of contents  after they are retrieved from the server
```
markcowl, line 12:
```suggestion
    /// Returns an AsyncPageable that transforms each page of contents  after they are retrieved from the server
```

Azure.ResourceManager.Core/Adapters/PhWrappingAsyncPageable.cs
```diff
@@ -3,135 +3,58 @@
...
+    /// accorgin to the profived transformation function
```
markcowl, line 13:
```suggestion
    /// according to the profived transformation function
```

Azure.ResourceManager.Core/ArmOperation.cs
```diff
@@ -1,182 +1,35 @@
...
-
 namespace Azure.ResourceManager.Core
 {
     /// <summary>
-    ///     Abstract class for long-running or synchronous applications. If we want to add ARM-specific OM, this is where we
-    ///     would add it.
-    ///     We may need to add ARM-specific OM, as customers have asked for additional configurability over polling stratgies
-    ///     on a per-operation basis
-    ///     TODO: GENERATOR Remove protected properties, as this should be integrated into the generated client
+    /// Abstract class for long-running or synchronous applications. If we want to add ARM-specific OM, this is where we
```
markcowl, line 7: This should be boiler plate text from the ```Operation<T>``` definition

markcowl (changes requested):
Some minor textual corrections - no re-review should be needed.

markcowl (approved):

bquantump (approved):

Issue #88, opened 2020-12-23
ArmResponse does not fill in Sku data of returned Azure.ResourceManager.Resources.Models.GenericResource


nisha-bhatia:
![image](https://user-images.githubusercontent.com/67986960/102945886-e4791380-4473-11eb-91ad-01a0344489dc.png)
![image](https://user-images.githubusercontent.com/67986960/102945927-feb2f180-4473-11eb-8c42-42414f35f5d1.png)

PR #132, opened 2021-01-16
Fix bug in ArmResponse of returned GenericResource

nisha-bhatia:
ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5081
Github issue: https://github.com/m-nash/azure-proto-sdk/issues/88

### DIFF
--- a/Azure.ResourceManager.Core/Resources/Plan.cs
+++ b/Azure.ResourceManager.Core/Resources/Plan.cs
@@ -10,6 +10,8 @@ namespace Azure.ResourceManager.Core
     /// </summary>
     public class Plan : IEquatable<Plan>, IComparable<Plan>
     {
+        private ResourceManager.Resources.Models.Plan plan;
+
         internal Plan(string name, string publisher, string product, string promotionCode, string version)
         {
             Name = name;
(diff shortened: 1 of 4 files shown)


Azure.ResourceManager.Core/Resources/Plan.cs
```diff
@@ -19,6 +21,16 @@ internal Plan(string name, string publisher, string product, string promotionCod
             Version = version;
         }
 
+        // TODO: Will be deleted after we move in azure core repo
+        internal Plan(ResourceManager.Resources.Models.Plan plan)
+        {
```
bquantump, line 26: Just call the other Constructor instead of setting one by one here

Azure.ResourceManager.Core/Resources/Sku.cs
```diff
@@ -19,6 +20,16 @@ internal Sku(string name, string tier, string family, string size, long? capacit
             Capacity = capacity;
         }
 
+        // TODO: Will be deleted after we move in azure core repo
+        internal Sku(Azure.ResourceManager.Resources.Models.Sku sku)
```
bquantump, line 24: Same as above, call other constructor instead of setting one by one

bquantump (approved):

m-nash (approved):

PR #133, opened 2021-01-19
Fill in comparison methods and unit tests for ResourceType

HarveyLink:
ADO Work Item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5004

### DIFF
--- a/Azure.ResourceManager.Core.Tests/ResourceTypeTests.cs
+++ b/Azure.ResourceManager.Core.Tests/ResourceTypeTests.cs
@@ -96,7 +96,6 @@ namespace Azure.ResourceManager.Core.Tests
             Assert.AreEqual(1, resourceType1.CompareTo(resourceType2));
         }
 
-
         [TestCase(-1, "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRg/providers/Microsoft.Network1/virtualNetworks2/Testvnet/subnets/default1",
             "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRg/providers/Microsoft.network2/virtualNetworks1/Testvnet/Subnets/default2")]
         [TestCase(1, "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRg/providers/Microsoft.network2/VirtualNetworks2/Testvnet/subnets2/default1",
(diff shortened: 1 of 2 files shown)


Azure.ResourceManager.Core.Tests/ResourceTypeTests.cs
```diff
@@ -0,0 +1,90 @@
...
+
+using NUnit.Framework;
+
+namespace Azure.ResourceManager.Core.Tests
+{
+    public class ResourceTypeTests
+    {
+        const string resourcePrefix = "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/testRg/providers/";
+
+        [TestCase(0, "Microsoft.ClassicStorage/storageAccounts/account1",
```
m-nash, line 12: add a test to compare to null

Azure.ResourceManager.Core.Tests/ResourceTypeTests.cs
```diff
@@ -0,0 +1,90 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using NUnit.Framework;
+
+namespace Azure.ResourceManager.Core.Tests
+{
+    public class ResourceTypeTests
```
m-nash, line 8: Some of the tests should pass in just the resourceType instead of the entire id, we should also compare the combinations.  Right now we are only comparing full id to full id
id.CompareTo(resourceType)
resourceType.CompareTo(id)
resourceType.CompareTo(resourceType)

m-nash (changes requested):

Azure.ResourceManager.Core.Tests/ResourceTypeTests.cs
HarveyLink: Since parameter null will throw ArgumentOutOfRangeException exception, I add object null compare here.

Azure.ResourceManager.Core/Resources/ResourceType.cs
```diff
@@ -87,7 +87,7 @@ public ResourceType Parent
 
         public int CompareTo(ResourceType other)
         {
-            if (other == null)
+            if ((object)other == null)
```
m-nash, line 90: typically you use reference equals here instead of a cast
```suggestion
            if (ReferenceEquals(other, null))
```

m-nash (changes requested):

m-nash (approved):

PR #136, opened 2021-01-22
Virtual Machine Add Tag

allenjzhang:


azure-proto-compute/VirtualMachineOperations.cs
```diff
@@ -286,5 +290,16 @@ public async Task<ArmOperation<VirtualMachine>> StartAddTagAsync(string key, str
                     return new VirtualMachine(ClientOptions, Resource as VirtualMachineData);
                 });
         }
+
+        private protected virtual VirtualMachine GetResource()
```
m-nash, line 294: Can we move this to ResourceOperationBase since all resource types will need this?

client/Scenarios/SetTagsOnVm.cs
```diff
@@ -11,12 +15,35 @@ public override void Execute()
...
             var rgOp = new AzureResourceManagerClient().ResourceGroup(Context.SubscriptionId, Context.RgName);
             var vmOp = rgOp.VirtualMachine(Context.VmName);
 
-            var vm = vmOp.Get().Value;
-            Console.WriteLine($"Adding tags to {vm.Data.Name}");
-            vm.StartAddTag("tagkey", "tagvalue");
+            Console.WriteLine($"Adding tags to {vmOp.Id.Name}");
+
+            var vm = (await vmOp.StartAddTag("key1", "value1").WaitForCompletionAsync()).Value;
+            DumpDictionary(vm.Data.Tags);
```
m-nash, line 30: Can we add an assert so it fails if the results isnt correct?

m-nash (changes requested):

azure-proto-compute/VirtualMachine.cs
```diff
@@ -18,9 +24,16 @@ internal VirtualMachine(ResourceOperationsBase operations, VirtualMachineData re
             Data = resource;
         }
 
-        /// <summary>
-        /// Gets the data representing this VirtualMachine.
-        /// </summary>
-        public VirtualMachineData Data { get; private set; }
+        /// <inheritdoc />
+        protected override VirtualMachine GetResource()
```
m-nash, line 28: I think we need to add this overload to all [Resource].cs classes otherwise StartAddTag for all other types will make an extra network call that isn't needed

azure-proto-compute/VirtualMachine.cs
```diff
@@ -18,9 +24,16 @@ internal VirtualMachine(ResourceOperationsBase operations, VirtualMachineData re
...
+        {
+            return this;
+        }
+
+        /// <inheritdoc />
+        protected override Task<VirtualMachine> GetResourceAsync()
```
m-nash, line 34: I think we need to add this overload to all [Resource].cs classes otherwise StartAddTagAsync for all other types will make an extra network call that isn't needed

m-nash (changes requested):

m-nash (approved):

PR #139, opened 2021-01-27
 Add proper create of rg with construct and fix null reference for add tags

bquantump:


Azure.ResourceManager.Core/ResourceGroupContainer.cs
```diff
@@ -32,17 +33,16 @@ internal ResourceGroupContainer(SubscriptionOperations subscription)
...
+        /// <param name="location"> The location of the resource group. </param>
+        /// <returns> A builder with <see cref="ResourceGroup"> and <see cref="ResourceGroupData"/>. </returns>
+        public ArmBuilder<ResourceGroup, ResourceGroupData> Construct(Location location, IDictionary<string, string> tags = default, string managedBy = default)
         {
-            var model = new ResourceGroupData(new ResourceManager.Resources.Models.ResourceGroup(location));
-            return new PhArmOperation<ResourceGroup, Azure.ResourceManager.Resources.Models.ResourceGroup>(
-                Operations.CreateOrUpdate(name, model),
-                g => new ResourceGroup(Parent, new ResourceGroupData(g)));
+            var model = new ResourceManager.Resources.Models.ResourceGroup(location);
+            model.Tags = tags;
```
m-nash, line 43: thought tags was unassignable, how does this work?

m-nash (changes requested):

Azure.ResourceManager.Core/ResourceGroupContainer.cs
bquantump: Apparently for resource group it is?

Is that a mistake?

m-nash: could be old library reference

client/Scenarios/ShutdownVmsByTag.cs
```diff
@@ -39,6 +39,8 @@ public override void Execute()
                 Console.WriteLine("--------Starting VM {0}--------", vm.Id.Name);
                 vm.PowerOn();
             }
+
```
m-nash, line 42: you have extra lines like this added in many a few files, should remove

azure-proto-compute/azure-proto-compute.csproj
```diff
@@ -6,6 +6,10 @@
     <LangVersion>latest</LangVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DocumentationFile>C:\Git\azure-proto-sdk\azure-proto-compute\obj\azure-proto-compute.xml</DocumentationFile>
```
allenjzhang, line 10: nit: Remove.

azure-proto-compute/VirtualMachineContainer.cs
```diff
@@ -101,21 +103,21 @@ public async override Task<ArmOperation<VirtualMachine>> StartCreateAsync(string
...
-        /// <param name="vmName"> The name of the Virtual Machine. </param>
+        /// <param name="hostName"> The hostname for the virtual machine. </param>
         /// <param name="adminUser"> The admin username to use. </param>
         /// <param name="adminPw"> The admin password to use. </param>
         /// <param name="nicId"> The network interface id to use. </param>
         /// <param name="aset"> The availability set to use. </param>
         /// <param name="location"> The location to create the Virtual Machine. </param>
         /// <returns> Object used to create a <see cref="VirtualMachine"/>. </returns>
-        public VirtualMachineModelBuilder Construct(string vmName, string adminUser, string adminPw, ResourceIdentifier nicId, AvailabilitySetData aset, LocationData location = null)
+        public VirtualMachineModelBuilder Construct(string hostName, string adminUser, string adminPw, ResourceIdentifier nicId, AvailabilitySetData aset, LocationData location = null)
```
allenjzhang, line 113: we should not using abbreviations on parameter names. Also would like to see we use ResourceIdentifier for AvalabilitySet as well.

azure-proto-compute/VirtualMachineContainer.cs
```diff
@@ -101,21 +103,21 @@ public async override Task<ArmOperation<VirtualMachine>> StartCreateAsync(string
         /// <summary>
         /// Construct an object used to create a VirtualMachine.
         /// </summary>
-        /// <param name="vmName"> The name of the Virtual Machine. </param>
+        /// <param name="hostName"> The hostname for the virtual machine. </param>
         /// <param name="adminUser"> The admin username to use. </param>
         /// <param name="adminPw"> The admin password to use. </param>
         /// <param name="nicId"> The network interface id to use. </param>
         /// <param name="aset"> The availability set to use. </param>
         /// <param name="location"> The location to create the Virtual Machine. </param>
```
allenjzhang, line 111: When location = null. We should clearly indicate where the default value will be coming from.

azure-proto-network/azure-proto-network.csproj
```diff
@@ -5,6 +5,10 @@
     <RootNamespace>azure_proto_network</RootNamespace>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <DocumentationFile>C:\Git\azure-proto-sdk\azure-proto-network\obj\azure-proto-network.xml</DocumentationFile>
```
allenjzhang, line 9: same. Did VSCode added this?

allenjzhang:
Steve, kudos for being thorough on updating the xml comments!!

Azure.ResourceManager.Core/ResourceGroupContainer.cs
bquantump: Should we changes this then?

bquantump:
> id prefer to not make this async.
> 
> To do this in scenarios override the normal void Execute() but in there call a private async Task ExecuteAsync()

Fixed

azure-proto-network/SubnetBuilder.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// A class representing a builder object to help create a subnet.
+    /// </summary>
+    public class SubnetBuilder : ArmBuilder<Subnet, SubnetData>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SubnetBuilder"/> class.
+        /// </summary>
+        /// <param name="container"> The container to create the subnet in. </param>
+        /// <param name="subnet"> The data model representing the subnet to create. </param>
+        public SubnetBuilder(SubnetContainer container, SubnetData subnet)
```
allenjzhang, line 15: This is to be public? Is this to be replicated across other resources? Maybe I missed the discussion to add this. So now we have different ways for construction:
1.  new XXXData();  -> XXXContainer.Create()
2.  XXXContainer.Construct().Create()
3.  new XXXBuilder(), ->  builder.Create()

azure-proto-compute/azure-proto-compute.csproj
m-nash: this is for the workshop to produce the xml docs

azure-proto-compute/VirtualMachineContainer.cs
m-nash: this has an upcoming ado item to track https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5333

azure-proto-network/NetworkSecurityGroupContainer.cs
```diff
@@ -82,17 +84,14 @@ public override async Task<ArmOperation<NetworkSecurityGroup>> StartCreateAsync(
         /// <summary>
         /// Construct an object used to create a network security group.
         /// </summary>
-        /// <param name="name"> The name of the network security group. </param>
-        /// <param name="location"> The location to create the network security group. </param>
+        /// <param name="locationData"> The location to create the network security group. </param>
         /// <param name="openPorts"> The location to create the network security group. </param>
         /// <returns> Object used to create a <see cref="NetworkSecurityGroup"/>. </returns>
-        public ArmBuilder<NetworkSecurityGroup, NetworkSecurityGroupData> Construct(string name, LocationData location = null, params int[] openPorts)
```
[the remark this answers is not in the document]
m-nash: yes hostname is the machines name vmname is the azure identifier for the vm.

yes we are removing this because of the duplicate construct vs create

azure-proto-network/SubnetBuilder.cs
m-nash: good catch making it internal

m-nash (approved):

PR #141, opened 2021-02-07
Adding sealed keyword for core classes + minor access modifier + typo…

allenjzhang:
… fixes


allenjzhang:
@m-nash , I added `sealed` to ResourceIdentifier.cs. If the thinking was for it to subclass into each RP Identifier, please let me know, and let's discuss.

m-nash:
Looks good other than the merge conflict

PR #150, opened 2021-02-18
Mnash 5332 update default subscription

m-nash:


Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -179,43 +166,12 @@ public T GetResourceOperations<T>(string subscription, string resourceGroup, str
...
-        /// <summary>
-        /// Gets default subscription.
-        /// </summary>
-        /// <param name="cancellationToken"> A token to allow the caller to cancel the call to the service. The default value is <see cref="P:System.Threading.CancellationToken.None" />. </param>
-        /// <returns> A <see cref="Task"/> that on completion returns the subscription id. </returns>
-        internal async Task<Subscription> GetDefaultSubscriptionAsync(CancellationToken cancellationToken = default)
+        private Subscription GetDefaultSubscription()
         {
-            return await GetSubscriptionContainer().GetDefaultSubscriptionAsync(cancellationToken);
+            var sub = GetSubscriptionContainer().List().FirstOrDefault();
```
bquantump, line 171: So is default subscription the first element of the result?

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
```diff
@@ -179,43 +166,12 @@ public T GetResourceOperations<T>(string subscription, string resourceGroup, str
...
+            if (sub is null)
```
bquantump, line 172: using "is" to avoid using "==" due to overload?

bquantump (approved):

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
m-nash: yes it returns in a repeatable pattern

Azure.ResourceManager.Core/AzureResourceManagerClient.cs
m-nash: yes

PR #153, opened 2021-02-19
Update Param Checking for SubscriptionOperations.cs

nisha-bhatia:


nisha-bhatia:
ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5142

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -57,8 +58,27 @@ protected SubscriptionOperations(SubscriptionOperations subscription, ResourceId
         /// </summary>
         /// <param name="resourceGroupName"> The name of the resource group. </param>
         /// <returns> The resource group operations. </returns>
+        /// <exception cref="ArgumentNullException"> ResourceGroupName cannot be null or empty. </exception>
+        /// <exception cref="ArgumentException"> ResourceGroupName cannot be longer than 90 characters. </exception>
+        /// <exception cref="ArgumentException"> ResourceGroupName cannot have a special character. </exception>
         public ResourceGroupOperations GetResourceGroupOperations(string resourceGroupName)
         {
+            if (string.IsNullOrWhiteSpace(resourceGroupName))
```
m-nash, line 66: I actually prefer if the validation of what a good resourceGroupName is belonged to the resourceGroupOperations class instead.  Your tests will still work as they are written but the exceptions would be thrown from the constructor of ResourceGroupOperations

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -57,8 +58,27 @@ protected SubscriptionOperations(SubscriptionOperations subscription, ResourceId
...
+            {
+                throw new ArgumentNullException("ResourceGroupName cannot be null or empty.");
+            }
+
+            if (resourceGroupName.Length > 90)
```
m-nash, line 71: when doing a single if(foo) throw you don't need the { }

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -57,8 +58,27 @@ protected SubscriptionOperations(SubscriptionOperations subscription, ResourceId
...
+            {
+                throw new ArgumentException("ResourceGroupName cannot be longer than 90 characters.");
+            }
+
+            Regex pattern = new Regex(@"^[-\w\._\(\)]+$");
```
m-nash, line 76: make this regex a const string inside resourcegroupoperations

m-nash (changes requested):

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -27,11 +28,20 @@ public class ResourceGroupOperations : ResourceOperationsBase<ResourceGroup>,
         /// </summary>
         /// <param name="options"> The client parameters to use in these operations. </param>
         /// <param name="rgName"> The name of the resource group to use. </param>
+        /// <exception cref="ArgumentException"> rgName cannot be longer than 90 characters. </exception>
```
m-nash, line 31: move these to the public methods that call this otherwise customers won't see the hints

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -27,11 +28,20 @@ public class ResourceGroupOperations : ResourceOperationsBase<ResourceGroup>,
...
             : base(options, $"{options.Id}/resourceGroups/{rgName}")
         {
+            if (rgName.Length > 90)
+                throw new ArgumentException("ResourceGroupName cannot be longer than 90 characters.");
+
+            if (!ValidationPattern.IsMatch(rgName))
+                throw new ArgumentException("ResourceGroupName cannot have a special character.");
         }
 
+        private Regex ValidationPattern = new Regex(@"^[-\w\._\(\)]+$");
```
m-nash, line 43: make this static readonly

m-nash (changes requested):

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -30,8 +31,15 @@ public class ResourceGroupOperations : ResourceOperationsBase<ResourceGroup>,
         internal ResourceGroupOperations(SubscriptionOperations options, string rgName)
             : base(options, $"{options.Id}/resourceGroups/{rgName}")
         {
+            if (rgName.Length > 90)
+                throw new ArgumentOutOfRangeException("ResourceGroupName cannot be longer than 90 characters.");
+
+            if (!ValidationPattern.IsMatch(rgName))
+                throw new ArgumentException("ResourceGroupName cannot have a special character.");
```
m-nash, line 38: we might just print the regex here instead of saying special characters because it can't contain whitespace either

Azure.ResourceManager.Core/SubscriptionOperations.cs
```diff
@@ -57,6 +58,8 @@ protected SubscriptionOperations(SubscriptionOperations subscription, ResourceId
         /// </summary>
         /// <param name="resourceGroupName"> The name of the resource group. </param>
         /// <returns> The resource group operations. </returns>
+        /// <exception cref="ArgumentOutOfRangeException"> resourceGroupName must be at least one character long and cannot be longer than 90 characters. </exception>
+        /// <exception cref="ArgumentException"> resourceGroupName cannot have a special character. </exception>
```
m-nash, line 62: same comment on the pattern

m-nash (changes requested):

m-nash (approved):

PR #156, opened 2021-02-23
Add Null Param Checking to Extension Methods

nisha-bhatia:


azure-proto-compute/Extensions/ArmClientExtensions.cs
```diff
@@ -18,12 +18,14 @@ public static class ArmClientExtensions
         /// <param name="resourceId"> The ResourceIdentifier of the resource that is the target of operations. </param>
         /// <returns> Returns an object representing the operations that can be performed over a specific <see cref="AvailabilitySet" />. </returns>
         /// <exception cref="ArgumentException"> ResourceIdentifier provided is not for an AvailabilitySet. </exception>
+        /// <exception cref="ArgumentNullException"> ResourceIdentifier cannot be null. </exception>
         public static AvailabilitySetOperations GetAvailabilitySetOperations(this AzureResourceManagerClient client, ResourceIdentifier resourceId)
         {
+            if (resourceId == null)
```
m-nash, line 24: use (resourceId is null)

azure-proto-compute/Extensions/ArmClientExtensions.cs
```diff
@@ -18,12 +18,14 @@ public static class ArmClientExtensions
...
+                throw new ArgumentNullException("resourceId cannot be null.");
```
m-nash, line 25: pass in the nameof(resourceId) instead of hard coding the name of the parameter

also double check your inputs, this takes in param name and message, make sure you are passing the right thing into the right place

azure-proto-compute/Extensions/ResourceGroupExtensions.cs
```diff
@@ -14,8 +15,11 @@ public static class ResourceGroupExtensions
         /// <param name="resourceGroup"> The <see cref="ResourceGroupOperations" /> instance the method will execute against. </param>
         /// <param name="vmName"> The name of the VirtualMachine. </param>
         /// <returns> Returns an object representing the operations that can be performed over a specific <see cref="VirtualMachine" />.</returns>
+        /// <exception cref="ArgumentNullException"> vmName cannot be null. </exception>
         public static VirtualMachineOperations GetVirtualMachineOperations(this ResourceGroupOperations resourceGroup, string vmName)
         {
+            if (string.IsNullOrWhiteSpace(vmName))
+                throw new ArgumentNullException("vmName cannot be null.");
```
m-nash, line 22: this would be an ArgumentException since you don't know if its null or whitespace

same comment on nameof(vmName) as above

m-nash (changes requested):

m-nash (approved):

PR #158, opened 2021-02-25
Fix potential bug

bquantump:
https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_workitems/edit/5420


Azure.ResourceManager.Core/ResourceListOperations.cs
```diff
@@ -39,6 +42,43 @@ public class ResourceListOperations
                 cancellationToken);
         }
 
+        /// <summary>
```
m-nash, line 45: This class is being phased out so I don't want to put extra stuff here.

m-nash (changes requested):

m-nash (approved):

PR #160, opened 2021-02-26
Add Param Checking to ResourceGroupOperations.cs

nisha-bhatia:
ADO item: https://dev.azure.com/azure-mgmt-ex/DotNET%20Management%20SDK/_boards/board/t/DotNET%20Management%20SDK%20Team/Stories/?workitem=5142

### DIFF
--- a/Azure.ResourceManager.Core/ResourceGroupOperations.cs
+++ b/Azure.ResourceManager.Core/ResourceGroupOperations.cs
@@ -130,13 +130,11 @@ namespace Azure.ResourceManager.Core
         /// <param name="key"> The key for the tag. </param>
         /// <param name="value"> The value for the tag. </param>
         /// <returns> A response with the <see cref="ArmOperation{ResourceGroup}"/> operation for this resource. </returns>
-        /// <exception cref="ArgumentException"> Key or value cannot be null. </exception>
+        /// <exception cref="ArgumentException"> Key cannot be null or a whitespace. </exception>
         public ArmResponse<ResourceGroup> AddTag(string key, string value)
         {
             if (string.IsNullOrWhiteSpace(key))
                 throw new ArgumentException(nameof(key), $"{nameof(key)} provided cannot be null or a whitespace.");
-            if (string.IsNullOrWhiteSpace(value))
-                throw new ArgumentException(nameof(value), $"{nameof(value)} provided cannot be null or a whitespace.");
 
             var resource = GetResource();
             var patch = new ResourceGroupPatchable() { Tags = resource.Data.Tags };
@@ -160,13 +158,11 @@ namespace Azure.ResourceManager.Core
         /// <param name="value"> The value for the tag. </param>
         /// <param name="cancellationToken"> A token to allow the caller to cancel the call to the service. The default value is <see cref="P:System.Threading.CancellationToken.None" />. </param>
         /// <returns> A <see cref="Task"/> that on completion returns a response with the <see cref="ArmOperation{ResourceGroup}"/> operation for this resource. </returns>
-        /// <exception cref="ArgumentException"> Key or value cannot be null. </exception>
+        /// <exception cref="ArgumentException"> Key cannot be null or a whitespace. </exception>
         public async Task<ArmResponse<ResourceGroup>> AddTagAsync(string key, string value, CancellationToken cancellationToken = default)
         {
             if (string.IsNullOrWhiteSpace(key))
                 throw new ArgumentException(nameof(key), $"{nameof(key)} provided cannot be null or a whitespace.");
-            if (string.IsNullOrWhiteSpace(value))
-                throw new ArgumentException(nameof(value), $"{nameof(value)} provided cannot be null or a whitespace.");
 
             var resource = GetResource();
             var patch = new ResourceGroupPatchable() { Tags = resource.Data.Tags };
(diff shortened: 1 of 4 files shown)


Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -129,8 +130,14 @@ public override async Task<ArmResponse<ResourceGroup>> GetAsync(CancellationToke
         /// <param name="key"> The key for the tag. </param>
         /// <param name="value"> The value for the tag. </param>
         /// <returns> A response with the <see cref="ArmOperation{ResourceGroup}"/> operation for this resource. </returns>
+        /// <exception cref="ArgumentException"> Key or value cannot be null. </exception>
         public ArmResponse<ResourceGroup> AddTag(string key, string value)
         {
+            if (string.IsNullOrWhiteSpace(key))
+                throw new ArgumentException(nameof(key), $"{nameof(key)} provided cannot be null or a whitespace.");
+            if (string.IsNullOrWhiteSpace(value))
```
m-nash, line 138: value can be null

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -232,11 +257,15 @@ public async Task<ArmOperation<ResourceGroup>> StartAddTagAsync(string name, str
...
             where TResource : TrackedResource
             where TOperations : ResourceOperationsBase<TOperations>
             where TContainer : ResourceContainerBase<TOperations, TResource>
         {
+            if (string.IsNullOrWhiteSpace(name))
+                throw new ArgumentException(nameof(name), $"{nameof(name)} provided cannot be null or a whitespace.");
+
             var myResource = model as TrackedResource;
 
             if (myResource == null)
```
m-nash, line 271: do we have a test case that exercises this path?  I would imagine this would fail if this was ever true.

Also use (myResource is null)

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -265,11 +294,15 @@ public async Task<ArmOperation<ResourceGroup>> StartAddTagAsync(string name, str
...
```
m-nash, line 308: same comment as above

client/Scenarios/CheckResourceGroupOpsAsync.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+
+            // Create AvailabilitySet
+            Console.WriteLine("--------Create AvailabilitySet async--------");
+            var aset = (await (await resourceGroup.GetAvailabilitySetContainer().Construct("Aligned").StartCreateOrUpdateAsync(Context.VmName + "_aSet")).WaitForCompletionAsync()).Value;
+            var data = aset.Get().Value.Data;
+            try
+            {                
+                rgOps.CreateResource<AvailabilitySetContainer, AvailabilitySet, AvailabilitySetData>("", data, LocationData.Default);
+            }
```
m-nash, line 74: should throw on this line since if you make it here then you didn't get the expected exception

m-nash (changes requested):

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -179,8 +188,12 @@ public async Task<ArmResponse<ResourceGroup>> AddTagAsync(string key, string val
         /// <remarks>
         /// <see href="https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-longrunning">Details on long running operation object.</see>
         /// </remarks>
+        /// <exception cref="ArgumentException"> Name cannot be null or a whitespace. </exception>
         public ArmOperation<ResourceGroup> StartAddTag(string name, string value)
```
m-nash, line 192: change param name to key

Azure.ResourceManager.Core/ResourceGroupOperations.cs
```diff
@@ -206,8 +219,12 @@ public ArmOperation<ResourceGroup> StartAddTag(string name, string value)
         /// <remarks>
         /// <see href="https://azure.github.io/azure-sdk/dotnet_introduction.html#dotnet-longrunning">Details on long running operation object.</see>
         /// </remarks>
+        /// <exception cref="ArgumentException"> Name cannot be null or a whitespace. </exception>
         public async Task<ArmOperation<ResourceGroup>> StartAddTagAsync(string name, string value, CancellationToken cancellationToken = default)
```
m-nash, line 223: change param name to key

client/Scenarios/CheckResourceGroupOpsAsync.cs
```diff
@@ -0,0 +1,173 @@
...
+            // Create Resource Group
+            Console.WriteLine($"--------Start create group {Context.RgName}--------");
+            var resourceGroup = subscription.GetResourceGroupContainer().Construct(Context.Loc).CreateOrUpdate(Context.RgName).Value;
+            CleanUp.Add(resourceGroup.Id);
+            var rgOps = subscription.GetResourceGroupOperations(Context.RgName);
+
+            try 
+            {
+                rgOps.AddTag("", "");
+            }
```
m-nash, line 34: you don't expect to make it to the next line, so you should throw right after line 33.

Also you can extract this into a method, take a lambda, and a message and exception type, will make this test much more readable

m-nash (changes requested):

m-nash (approved):

